PR #111, opened 2015-10-20
Feature/install examples

andre-merzky:
Hi Vivek -- this PR is supposed to install the repex `examples/` subtree under `$VE/share/radical.ensemblemd/examples/`.  Please check carefully that it does not break install otherwise before merge!

The purpose is that, during the tutorial, people don't need to run `wget` or `scp` to get the examples onto the workflow machine...


andre-merzky:
PS.: I am not sure if I created that PR toward the right branch -- let me know if it should point elsewhere, and I'll recreate it!

Issue #113, opened 2015-10-21
pipeline example waits for more units than it submits?


andre-merzky:
the output shows:

```
submit 16 unit(s)
        ................                                                      ok
wait for 16 unit(s)
        ++++++++++++++++                                                      ok
                                                                            done
Waiting for step_2 to complete.                                                \
submit 16 unit(s)
        ................                                                      ok
wait for 32 unit(s)
        ++++++++++++++++++++++++++++++++                                      ok
```

so the second step submits 16 units, but waits for 32?  I assume it waits for the units from step_1 again -- but that is not very intuitive.  Also, not needed...

vivek-bala:
I think that is from rp. The reporting seems to be cumulative.

andre-merzky:
`wait for 32 unit(s)` - it is not just the reporting: it really waits for that many units!  I assume (but did not check) that enmd simply calls `umgr.wait_units()`?  That would wait for all units ever submitted to that umgr -- which explains the numbers.  You probably want to pass a list of unit IDs: `umgr.wait_units(unit_ids_for_this_step)` ...

andre-merzky:
PS.: don't bother fixing this for the tutorial...

vivek-bala:
https://github.com/radical-cybertools/radical.ensemblemd/blob/master/src/radical/ensemblemd/exec_plugins/pipeline/static.py#L332

Every step has "N" CUs and every step waits for the "N" CUs to finish. So I
submit N at a time and I do wait_units(), doesn't that mean you wait for N
CUs at a time (since the N from the previous step have already finished -
"Done"). In the above case N=16.

andre-merzky:
`wait_units()` will wait on *all* units which have ever been submitted to that umgr.  How would the umgr know which ones *not* to wait for?

If you want  to wait for N CUs, then you need to pass the UIDs for those N CUs to `wait_units()`...

vivek-bala:
> wait_units() will wait on all units which have ever been submitted to that umgr. How would the umgr know which ones not to wait for?

But the first N CUs have finished executing (reach 'Done' state). I cannot understand "waiting" for those completed CUs.

andre-merzky:
consider (pseudo code):

```
umgr = UnitManager()
unit_1 = umgr,submit('sleep 1')
sleep(5) # unit 1 is DONE now
unit_2 = umgr,submit('sleep 1')
umgr.wait_units()
```

[...]

PR #119, opened 2015-10-22
fix unit reporting/waiting

andre-merzky:
This addresses #113, and also makes the `wait_units()` calls somewhat more efficient...


vivek-bala:
Thanks andre !

PR #125, opened 2015-12-15
environment takes a dict of environment variables + values

dotsdl:
fix to doc example for kernel specification

The docs currently say that environment variables are set with a list, but instances of `Kernel`s defined in the library use a dict here, which makes more sense anyway.


andre-merzky:
Woah, a pull request for documentation!  Niiice!! :D

dotsdl:
It takes a village. :D

vivek-bala:
Indeed, it should be a dict. Thanks for the fix :)

PR #224, opened 2018-04-10
json data format

jdakka:
* Modified `appmanager.py` to collect data about `PST` and `resource dictionary` (for now)

*`output.json` takes the form of `entk_self.uid.json` 

* content of mock file (in this case I did not call `.run()` so you won't see output of resource dict:

```
{"pipelines": [{"Total pipeline(s)": 1}, {"pipeline": 0, "pipelines uid": "radical.entk.pipeline.0000", "pipeline name": "", "total stages": 1}, {"stage uid": "radical.entk.stage.0000", "stage name": "", "total tasks": 1, "stage": 0}, {"task uid": "radical.entk.task.0000", "task": 0, "task name": "my-first-task"}], "dictionary": []}
```

I'll continue to iterate on the necessary components but this is an update


src/radical/entk/appman/appmanager.py
```diff
@@ -129,7 +136,8 @@ def resource_manager(self):
         :getter: Returns the resource manager object being used
         :setter: Assigns a resource manager
         """
-
+    
+        pprint.pprint(self._resource_manager)
         return self._resource_manager
```
andre-merzky: This is probably a debug print and should be removed, or converted to a log statement?

src/radical/entk/appman/appmanager.py
```diff
@@ -173,6 +181,47 @@ def assign_workflow(self, workflow):
...
+        self._data['pipelines'].append({'Total pipeline(s)': len(pipelines)})
+
+        for i in range(len(pipelines)):
+            pipeline = pipelines[i]
+            stages = list(pipeline.stages)
+
+            self._data['pipelines'].append({'pipeline': i, 
+                                      'pipeline name' : pipeline.name,
+                                      'pipelines uid' : pipeline.uid, 
+                                      'total stages': len(stages)})
```
andre-merzky: Why do you record `i`?  Also, this is a pipeline entry,  and a shorter dict would suffice:
```
{'name'  : pipeline.name,
 'uid'   : pipeline.uid,
 'stages': len(stages)
}
```

Same holds for the other dicts below.

src/radical/entk/appman/appmanager.py
```diff
@@ -173,6 +181,47 @@ def assign_workflow(self, workflow):
...
+                    len(tasks)))
+                
+
+                for k in range(len(tasks)): 
+                    self._data['pipelines'].append({'task': k, 
+                                      'task name' : tasks[k].name,
+                                      'task uid': tasks[k].uid})
+                    pprint.pprint('Tasks: {0}'.format(tasks[k].name))
+
+        pprint.pprint(self._data)
```
andre-merzky: This is also a debug print I guess?  What about the other `pprint`s above?

src/radical/entk/appman/appmanager.py
```diff
@@ -173,6 +181,47 @@ def assign_workflow(self, workflow):
...
+
+                for k in range(len(tasks)): 
+                    self._data['pipelines'].append({'task': k, 
+                                      'task name' : tasks[k].name,
+                                      'task uid': tasks[k].uid})
+                    pprint.pprint('Tasks: {0}'.format(tasks[k].name))
+
+        pprint.pprint(self._data) 
+        with open('entk_{}.json'.format(self._uid), 'w') as outfile:  
+            json.dump(self._data, outfile)
```
andre-merzky: please use 
```
ru.write_json(self._data, 'entk_%s.json' % self._uid)
```

src/radical/entk/appman/appmanager.py
```diff
@@ -203,7 +252,11 @@ def run(self):
 
             else:
 
+                with open('entk_{}.json'.format(self._uid), 'w') as outfile:  
+                    json.dump(self._data['resource dictionary'].append(self._resource_manager), outfile)
+
```
andre-merzky: Please use `ru.write_json(data, filename)`

src/radical/entk/appman/appmanager.py
```diff
@@ -203,7 +252,11 @@ def run(self):
...
                 self._prof.prof('amgr run started', uid=self._uid)
+                self._resource_manager
```
andre-merzky: The fact that you use an empty attribute statement here suggests that the json dump should be moved to a separate, private (or public?) method.  Attribute statements are for inspection, and should have minimal side effects.

jdakka:
I pushed my changes based on pull request reviews from @andre-merzky , can you please review.

src/radical/entk/appman/appmanager.py
andre-merzky: i still don't understand why this is implemented as a side effect to an attribute.

Consider the following (made-up) example:

```
while  self._resource_manager.check_state():
  print 'ok'
```
This would, at every iteration of that loop, save the json data set, fo no reason whatsoever.  The implementation basically hijacks the attribute.  Why not move the json dump to, well, something like `self._dump_json()` or whatever, which is called when the intent is to dump the json files?

vivek-bala:
Thanks Andre, Jumana. I have made a few changes to get more detail in the json, moved the function to another file and added a test case.

Issue #234, opened 2018-06-05
Adding data example has typo


iparask:
The `adding_data.py` example as well as the [documentation](http://radicalentk.readthedocs.io/en/latest/user_guide/adding_data.html) has a typo.

Instead of writing `$Pipeline_%s_Stage_%s_Task_%s/output.txt'%(p.uid, s1.uid, t1.uid)`, it writes: `$Pipline_%s_Stage_%s_Task_%s/output.txt'%(p.uid, s1.uid, t1.uid)`

vivek-bala:
Fixed in devel. Thanks!

PR #235, opened 2018-06-05
Change based on #234

iparask:


vivek-bala:
Didn't merge the PR but patched the diff in the last release.

PR #248, opened 2018-07-11
Feature/custom appname

iparask:
Test creation is pending


vivek-bala:
Hey Giannis, two things:

* Can you add a test in tests/test_component/test_amgr
* I noticed that the PR is towards master now, I would like to merge with master, run all the tests and then release it. Any particular reason the PR is set to master right now? If not, can you redirect the PR please?

Thanks!

vivek-bala:
I think you can change the target of a PR. See [here](https://stackoverflow.com/questions/24159036/how-to-modify-a-pull-request-on-github-to-change-target-branch-to-merge-into)

vivek-bala:
But feel free to delete and recreate if you think that'll be simpler.

vivek-bala:
Hey Giannis, how's the PR going? Were you able to work on the tests?

tests/test_component/test_amgr.py
```diff
@@ -16,6 +16,37 @@
 MLAB = 'mongodb://entk:entk123@ds143511.mlab.com:43511/entk_0_7_4_release'
 
 def test_amgr_initialization():
+    amgr_name = ru.generate_id('test.appmanager.%(item_counter)04d', ru.ID_CUSTOM)
+    amgr = Amgr(hostname=hostname, port=port,name=amgr_name)
+
+    assert amgr._name.split('.') == amgr_name.split('.')
+    assert amgr._sid.split('.') == amgr_name.split('.')
+    assert amgr._uid.split('.') == ['appmanager', '0000']
```
vivek-bala: I think line 24-49 is duplicated. Do you think this is necessary? 
IMO, specifying amgr name only effects amgr._name and amgr._sid, so lines 22-23 should 
suffice. What do you think?

vivek-bala:
The AppManager creates the Rmgr and populates it as in lines https://github.com/radical-cybertools/radical.entk/blob/feature/custom_appname/src/radical/entk/appman/appmanager.py#L197-L203. So it should be enough to test the populate method in the test_rmgr.py (which you already have).

But, I don't think you have to change any of the tests in tests/test_component/test_rmgr.py. You have essentially changed one string (i.e. the sid) to another. I think we can leave it as it was or change the sid everywhere (in all the files) to be consistent. I think the latter might not be the best use of time. What do you think? Let me know if you a specific reason to replace the sids in the tests.

tests/test_component/test_amgr.py
iparask: I think we need to make sure that `__init__` runs correctly whether the name is setup or not. I can remove them, though

iparask:
By my understanding, the resource manager (rmgr) is responsible to create the necessary folders, right? 

I replaced sid in rmgr tests because this setups up the folder name for EnTK's session and RP's session. We are starting to have the same folders, as a result, we need to make sure that this does not create an issue when profiles and logs are written.

If you think, it is unnecessary, I will remove them. 

I am moving forward with Integration testing.

iparask:
@vivek-bala ping

Issue #286, opened 2018-12-06
Invalid placeholder Error


iparask:
I am using the devel branch of EnTK and I am getting the following error:

```
2018-12-06 11:29:44,753: radical.entk.task_manager.0000: task-manager                    : umgr.0000.subscriber._state_sub_cb: DEBUG   : Attempting to sync task.0000 with state EXECUTED with AppManager
2018-12-06 11:29:44,768: radical.entk.task_manager.0000: task-manager                    : umgr.0000.subscriber._state_sub_cb: DEBUG   : task.0000 with state EXECUTED synced with AppManager
2018-12-06 11:29:44,770: radical.entk.task_manager.0000: task-manager                    : umgr.0000.subscriber._state_sub_cb: INFO    : Transition of task.0000 to new state EXECUTED successful
2018-12-06 11:29:44,773: radical.entk.task_manager.0000: task-manager                    : umgr.0000.subscriber._state_sub_cb: INFO    : Pushed task task.0000 with state EXECUTED to completed queue test_run_lfs_5-completedq-1
2018-12-06 11:29:44,843: radical.entk.task_manager.0000: task-manager                    : MainThread     : ERROR   : Error in task execution: Value for attribute placeholder of object length incorrect. Expected value $Pipeline_{pipeline.uid}_Stage_{stage.uid}_Task_{task.uid} or $SHARED, but got [u'$NODE', u'LFS', u'PATH'].
Traceback (most recent call last):
  File "/home/iparask/miniconda2/envs/EnTKlfs/lib/python2.7/site-packages/radical/entk/execman/rp/task_manager.py", line 205, in _tmgr
    bulk_cuds.append(create_cud_from_task(t, placeholder_dict, local_prof))
  File "/home/iparask/miniconda2/envs/EnTKlfs/lib/python2.7/site-packages/radical/entk/execman/rp/task_processor.py", line 356, in create_cud_from_task
    cud.arguments = resolve_arguments(task.arguments, placeholder_dict)
  File "/home/iparask/miniconda2/envs/EnTKlfs/lib/python2.7/site-packages/radical/entk/execman/rp/task_processor.py", line 107, in resolve_arguments
    actual_value=broken_placeholder)
ValueError: Value for attribute placeholder of object length incorrect. Expected value $Pipeline_{pipeline.uid}_Stage_{stage.uid}_Task_{task.uid} or $SHARED, but got [u'$NODE', u'LFS', u'PATH'].
[...]

PR #287, opened 2018-12-06
fix resolve_args + bring back resolve_tags + tests

vivek-bala:
This PR fixes https://github.com/radical-cybertools/radical.entk/issues/286. Does the following: 
* parsing the argument and only resolves if the string has '$SHARED' or '$Pipeline', ignores rest.
* brings back resolve_tags method that seems to have been lost in an earlier merge
* tests resolve_args (if it indeed ignores) and resolve_tags()


iparask:
Can we merge this?

vivek-bala:
Yup, doing it now. Give me a few minutes for the CI to pass.

PR #297, opened 2019-01-26
allow any callble for post_exec hooks

andre-merzky:
This enables the use of class methods and other closures to be used as callbacks.

I upped the  priority to `high` because this seems to simplify the repex use case.  But let me know if you think that it needs longer time to iterate...


vivek-bala:
Yup, checking for callable makes sense. I didn't realize instance methods don't get picked up. Can I also ask you to fix the tests in this PR or you want me to do it? :)

vivek-bala:
Can you change the base of the PR to devel though? I would like to merge to master after the tests have been modified.

andre-merzky:
Ah, sure - base is changed now.  You can select `devel` as default branch in the settings BTW, then all PRs got toward devel by default.  Sorry though for not paying attention...

andre-merzky:
I added tests and they seem to pass - but the test suite overall does not.  I am running it with `pytest tests/` and get `31 failed, 83 passed, 60 warnings`.  Any idea?  I id not look closer myself, yet...

vivek-bala:
I'll take care of the remaining tests as I am merging them.

PR #301, opened 2019-02-06
Fix/logger tgt

andre-merzky:
The hardcoded logger targets make it impossible to redirect logs to a file *and* to silence it on stdout.  The patch now enables the setting via `RADICAL_ENTK_LOG_TGT` which is constent with the remainder of the RCT stack.


mturilli:
Hi @andre-merzky, would you mind to review the failed tests in travis and see whether they depend on your PR?

andre-merzky:
> Hi @andre-merzky, would you mind to review the failed tests in travis and see whether they depend on your PR?

No, they fail for changes in the post-exec handling.  I have to admit though that this also was a PR of mine, so let me fix that ;-)  There are some other unrelated errors (I think), I will leave them alone.

andre-merzky:
I think I addressed some of the tests related to the post-exec PR.  Vivek, attached is an output of a test run.  Any advise on the failing tests?

[entk_tests.log](https://github.com/radical-cybertools/radical.entk/files/2847240/entk_tests.log)

andre-merzky:
Vivek, I just realized that the  log fix branch (the PR you just approved) was not branched off devel, but from the suspend/resume fix.   I thus rerouted this PR to fix/issue_296.  Apologies for that.

vivek-bala:
Do you think we include issue_296 branch for next month's release or keep it for later? I think this PR can be merged with devel and prepped for next month's release. What do you think?

andre-merzky:
Whom is this question for? :-)

vivek-bala:
Hehe, you indeed. I was wondering that since you rebased it from devel to issue_296. I can see if we can merge it with devel and release it next month.

andre-merzky:
:-)  Release schedule is up to you, mate!

vivek-bala:
Alrighty :)

mturilli:
The PR policy for EnTK (and it should really be for RCT as a whole) is to fix all the tests that the PR brakes. This is why we have Travis to run for each PR and why we cannot merge a PR before all tests pass. I see that `test_write_session_description` failing. Looking at the error, it might not be related to EnTK (seems t be originating from RU) but, even in that case, it would be a good indication that something is wrong with the stack. @andre-merzky, could you fix it and add any test for the next code you have added (if any)?

andre-merzky:
How can that test have ever worked?  It calls:
```
ru.write_json(desc, '%s/radical.entk.%s.json' % (amgr._sid, amgr._sid))
```

but ru.write_json never attempted to create a missing subdirectory, and that is what this test stumbles over.  What is the expected behavior: should that directory exist (from some other test or mockup), or is an explicit `mkdir` missing, or is RU expected to take care of this?

I don't think the test failure is related to the PR (or at least I don't yet see how).

iparask:
The `test_write_session_description` fails. I'll check when the test passes!

andre-merzky:
ping @vivek-bala: I am not sure how to handle that test failure - see comment above.  Any advise?

andre-merzky:
Can this be looked at, please?  Whenever I need some clean EnTK output when debugging, I have to apply this logger fix manually, and remove it again before committing.  It was originally a very trivial change of 3 lines of code, and now is stuck in discussion about unrelated changes and tests...

src/radical/entk/VERSION
```diff
@@ -1 +1 @@
-0.7.11-0.7.10-19-gfba5998@master
+0.7.11-0.7.11-2-ga85a1733@fix-logger_tgt
```
andre-merzky: Why is this file in the repo actually?  It gets auto-created by setup.py...

andre-merzky:
This is too far out of sync with the test changes in devel, partly because this PR attempted to fix tests whose code was not touched in the PR.  I'll redo that with the  original focus on the log target fixes.

Issue #306, opened 2019-02-15
Reporter to report parent stage and pipeline of tasks and stages


mturilli:
The default output of EnTK is something like:
```
EnTK session: re.session.Maras-MBP.furious.lan.mturilli.017942.0009
Creating AppManager                                                           ok
Validating and assigning resource manager                                     ok
Setting up RabbitMQ system                                                    ok
All components created
Update: Pipeline pipeline.0000 in state SCHEDULING
Update: Stage stage.0000 in state SCHEDULING
Update: Task task.0000 in state SCHEDULING
Update: Pipeline pipeline.0001 in state SCHEDULING
Update: Stage stage.0002 in state SCHEDULING
Update: Task task.0007 in state SCHEDULING
Update: Task task.0008 in state SCHEDULING
Update: Task task.0000 in state SCHEDULED
```
When prototyping a workflow skeleton, it is difficult to infer the workflow's structure from that output. Something like the following would make it easier:
```
Update: Task task.0000, Stage stage.0000, Pipelines pipeline.0000 in state SUBMITTING
```

vivek-bala:
I think we should capture the information that specified state is for a task/stage/pipeline.
My suggestion:

```
# For tasks
Update: Task task.0000 of Stage stage.0000 of Pipeline pipeline.0000 in state SUBMITTING
# For stages
Update: Stage stage.0000 of Pipeline pipeline.0000 in state SCHEDULING
# For pipelines
Update: Pipeline pipeline.0000 in state SCHEDULING
```

mturilli:
I tested #307 and it works. Unfortunately, it is less useful than I thought. I assigned names to pipelines and stages as per API spec but they are not reported by the reporter. Any insight in how workflow structure/priority checking is usually done in EnTK when testing a new workflow app?

andre-merzky:
There might not exist a single good solution, and I would caution against over-engineering this.  Having said that, we could switch behavior with an env variable and report names instead of UIDs, to get something like:
```
Update: pipeline.0009.stage.0029.task.0029 state: DEQUEUED
Update: pipeline.0009.stage.0029.task.0029 state: DONE
Update: pipeline.0009.stage.0029 state: DONE
Update: pipeline.0009 state: DONE
```

switched to 

```
Update: pipeline.0008.md_stage.task.0028 state: DEQUEUED
Update: pipeline.0008.md_stage.task.0028 state: DONE
Update: pipeline.0008.md_stage state: DONE
[...]

PR #307, opened 2019-02-15
implement long uids

andre-merzky:
This addresses #306:

```
...
Update: pipeline.0009.stage.0029.task.0029 state: DEQUEUED
Update: pipeline.0009.stage.0029.task.0029 state: DONE
Update: pipeline.0009.stage.0029 state: DONE
Update: pipeline.0009 state: DONE
...
...

### DIFF
--- a/src/radical/entk/appman/appmanager.py
+++ b/src/radical/entk/appman/appmanager.py
@@ -677,7 +677,7 @@ class AppManager(object):
 
                                             mq_channel.basic_ack(delivery_tag=method_frame.delivery_tag)
                                             self._report.ok('Update: ')
-                                            self._report.info('Task %s in state %s\n' % (task.uid, task.state))
+                                            self._report.info('%s state: %s\n' % (task.luid, task.state))
 
                                             found_task = True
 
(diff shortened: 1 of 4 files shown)


vivek-bala:
LGTM! I just rebased it towards devel (from master). Let me know if you wanted to move back to master (hotfix?).

andre-merzky:
@mturilli , @vivek-bala : So whats the merge policy on failing tests?

mturilli:
No merge with failure?

vivek-bala:
We rerun the tests till they pass :P

On a serious note, sometimes 1-2 tests fail because they don't get the expected state for a Task or the resource. This was the reason why initially the test failed for this PR. The solution so far has been to rerun the test suite (can be done directly on Travis).

If the test fails because of other errors (which can be seen by looking at the test results on Travis), I think we make the appropriate code changes or test changes to fix the failure and push back to the same PR.

andre-merzky:
So why do the tests fail sometimes?   Doesn't that defeat the purpose of running tests (a little)? ;-)

vivek-bala:
I should have phrased it better. What I have seen is that sometimes pilot doesn't report its final state when run on localhost. I agree there should be a test for that as well, but I think that would go in RP rather than here. Put in another way, I think that would be an RP test rather an EnTK-RP integration test. What do you think?

I should also say that I don't know how to reproduce the error consistently though.

andre-merzky:
Underdtood, thanks  Vivek!

PR #308, opened 2019-02-16
restructure app manager

andre-merzky:
This ended up to be a little more heavy-handed than I wanted - but I **really** wanted to get rid of the near-infinite nesting.  The previous structure was like this:

```python
def synchronizer(...):
    ...

    def update_pipe(msg, ...):
        send(..., tag)
    def update_stage(msg, ...):
        send(..., tag)
    def update_task(msg, ...):
        send(..., tag)

    msg = ...
    tag = ...
    task = ...

    for pipe in workflow:
        if not pipe.completed:
            if pipe has task.parent_stage:
                for stage in pipe.stages:
                    if not stage.completed:
                        if stage has task:
                            update_task(...)

    # similar for stage and pipeline updates
```

This is now closer to:
```python
def _update_pipe(self, msg, tag, ...):
        send(..., tag)
def _update_stage(self, msg, tag, ...):
        send(..., tag)
def _update_task(self, msg, tag, ...):
        send(..., tag)

def synchronizer(...):
    msg = ...
    tag = ...
    task = ...

    pipe = <find pipe in self.workflow>
    if pipe:
        stage = <find stage in pipe.stages>
    if stage:
        task = <find task in stage.tasks>
    if task:
        update_task(...)

    # similar for stage and pipeline updates
```

The `<find x in y>` parts are a little opaque, since it requires search in lists (that's what the nested for loops did before) - a dictionary by uid provide a more direct, faster and readable access:

```
def synchronizer(...):
    pipe = self.workflow.pipes.get(pipe_id)
    if pipe:
        stage = pipe.stages.get(stage_id)
    if stage:
        task = stage.tasks.get(task_id)
    if task:
        update_task(...)
```
But that was out of scope this time, and I would definitely need Vivek's feedback before touching that kind of data structure.

I did not add tests, since neither semantics nor API syntax changed (and the inner methods did not have unit tests to begin with, which makes sense, as they are really just for code encapsulation).  Lets see how the tests behave...

Feedback welcome!

### DIFF
--- a/src/radical/entk/appman/appmanager.py
+++ b/src/radical/entk/appman/appmanager.py
@@ -2,44 +2,57 @@ __copyright__ = "Copyright 2017-2018, http://radical.rutgers.edu"
 __author__ = "Vivek Balasubramanian <vivek.balasubramaniana@rutgers.edu>"
 __license__ = "MIT"
 
-import radical.utils as ru
-from radical.entk.exceptions import *
-from radical.entk.pipeline.pipeline import Pipeline
-from radical.entk.stage.stage import Stage
-from radical.entk.task.task import Task
-from radical.entk.utils.prof_utils import write_session_description
-from radical.entk.utils.prof_utils import write_workflow
-from wfprocessor import WFprocessor
 import time
 import os
-import Queue
 import pika
 import json
+
 from threading import Thread, Event
-from radical.entk import states
+
+import radical.utils as ru
+
+from ..exceptions        import *
+from ..pipeline.pipeline import Pipeline
+from ..stage.stage       import Stage
+from ..task.task         import Task
+from ..utils.prof_utils  import write_session_description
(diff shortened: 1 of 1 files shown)


andre-merzky:
@vivek-bala : I left a couple of comments marked with `AM` which point to code where I would like to get confirmation that semantics is maintained.  For example (and I forgot to mention that above), I found 3 code paths which seemed to do the essentially same thing during termination, and I collapsed those into a `_terminate()` method - but I might have missed some differences in those paths, since they expressed the termination sequence in slightly different syntax, and I could not discern why.  Thanks!

iparask:
@andre-merzky Travis fails with a couple of failing tests. Also there is a conflict with `devel`. Can you fix them before I continue here?

andre-merzky:
The conflict is basically all the content of the PR, and resolving that conflict does not make sense - this needs to be redone :/

PR #309, opened 2019-02-16
Tmp/andre setup

andre-merzky:
Sync `setup.py` with RU's `setup.py`.

The only semantic change is that all examples are installed (incl. `misc/`), and the removal of the additional `scripts/` subdir in each example dir.


setup.py
```diff
@@ -173,90 +206,99 @@ def visit((prefix, strip, found), dirname, names):
...
-    'keywords'         : "ensemble workflow execution",
-    'classifiers'      :  [
+    'name'               : name,
+    'version'            : version,
+    'description'        : "Radical Ensemble Toolkit.",
+    'long_description'   : (read('README.md') + '\n\n' + read('CHANGES.md')),
+    'author'             : 'RADICAL Group at Rutgers University',
+    'author_email'       : 'radical@rutgers.edu',
+    'maintainer'         : 'The RADICAL Group',
+    'maintainer_email'   : 'radical@rutgers.edu',
```
iparask: Does this email exist?

setup.py
```diff
@@ -173,90 +206,99 @@ def visit((prefix, strip, found), dirname, names):
...
+    'packages'           : find_packages('src'),
+    'package_dir'        : {'': 'src'},
+    'scripts'            : ['bin/entk-version'],
+    'package_data'       : {'': ['*.sh', '*.json',
+                                 'VERSION', 'SDIST', sdist_name]},
+    'setup_requires'     : ['pytest-runner'],
+    'install_requires'   : ['radical.utils',
+                            'radical.pilot',
+                            'pika',
+                            'sphinx'
```
iparask: I'm still not sure why `sphinx` is needed for installation. It is used for documentation generation.

src/radical/entk/__init__.py
```diff
@@ -1,9 +1,10 @@
...
-from radical.entk.version import version, __version__
+from .version import version, __version__, version_short, version_detail
+from .version import version_base, version_branch, sdist_name, sdist_path
 
-from radical.entk.pipeline.pipeline import Pipeline
-from radical.entk.stage.stage import Stage
-from radical.entk.task.task import Task
+from .pipeline.pipeline import Pipeline
+from .stage.stage       import Stage
+from .task.task         import Task
```
iparask: These imports are relative, but the one in line 9 is not. Shouldn't it be relative for consistency. Although, I think `pylint` and `pep8` complain with relative imports

src/radical/entk/pipeline/__init__.py
```diff
@@ -0,0 +1,3 @@
+
```
iparask: Can a banner be placed here?

setup.py
andre-merzky: Good point - I thought so, but we can also use a different one, like one of our lists.  What I do know is that this specific email has a finite lifetime :-)

setup.py
andre-merzky: Ack.

src/radical/entk/__init__.py
andre-merzky: Why do they complain?  
Yes, I will look at `states`.

setup.py
vivek-bala: Makes sense. Which one should we use? I would think maintainer email should be of an individual right or a group email (not rct-devel or something).

setup.py
vivek-bala: Ack. Will change.

src/radical/entk/pipeline/__init__.py
vivek-bala: What do you mean by a banner?

PR #316, opened 2019-02-26
avoid duplicated state history update

andre-merzky:


mturilli:
Hi @andre-merzky, this seems to be breaking `_state_history` in `test_pipeline_suspend_resume()`.

andre-merzky:
> Hi @andre-merzky, this seems to be breaking `_state_history` in `test_pipeline_suspend_resume()`.

Yeah, I am not surprised, really :-)  Lets see what @vivek-bala says about this.  There are three different ways used to set a state of a pipeline (that I can see): assign to `self._state`, use the `state` setter, and `transition()`.  I don't think they mean the same, semantically - and the failing test reflects that.

vivek-bala:
Hey Andre, every time you change the state I think it needs to be added to the state history list. Can you point me to the other line where SUSPEND is being added to the state history?

andre-merzky:
I think this PR may have been premature.  I observed `SUSPENDED` being added twice, but that might have been on a non-virgin code.  Apologies....

PR #317, opened 2019-02-27
Fix/examples 

jdakka:
changed task executables from list to str for examples scripts


andre-merzky:
Hey Jumana -  thanks for that!  But why was it closed?  Master does not yet have string execs - but that only means that the PR has to wait for a bit... :-)

vivek-bala:
Same here. @jdakka any particular why this PR was closed? #318 seems very similar to this one.

jdakka:
@vivek-bala Because you asked me to push to `devel` not `master`

vivek-bala:
Oh! I see. Thanks. You can use the edit button on the top right of the PR to change the target branch :)

jdakka:
nifty :)

Issue #322, opened 2019-03-01
radical.entk version 0.7.13 breaks execution of example scripts


uvaaland:
I have been running radical.entk version 0.7.11 for a couple of weeks without any problem on my virtual machine. However, recently when I helped someone else with the exact same setup, we ran into a problem when trying to run any of the example scripts.

Trying to run `get_started.py` or any of the other examples, we get the following error:

```
EnTK session: re.session.test.vagrant.017956.0004
Creating AppManager                                                           ok
Validating and assigning resource manager                                     ok
Setting up RabbitMQ system                                                    ok
All components created
Update: Pipeline pipeline.0000 in state SCHEDULING
Update: Stage stage.0000 in state SCHEDULING
Update: Task task.0000 in state SCHEDULING
Update: Task task.0000 in state SCHEDULED
Update: Stage stage.0000 in state SCHEDULED
/home/vagrant/scratch/entk/venv/local/lib/python2.7/site-packages/pymongo/topology.py:149: UserWarning: MongoClient opened before
fork. Create MongoClient only after forking. See PyMongo's documentation for details: http://api.mongodb.org/python/current/faq.ht
ml#is-pymongo-fork-safe
  "MongoClient opened before fork. Create MongoClient only "
Update: Task task.0000 in state SUBMITTING
Update: Task task.0000 in state SUBMITTED
2019-03-01 21:17:36,802: radical.entk.appmanager.0000: MainProcess                     : synchronizer-thread: ERROR   : Unknown e$
ror in synchronizer: Expected (base) type(s) <type 'str'>, but got <type 'list'>..
 Terminating thread
Traceback (most recent call last):
  File "/home/vagrant/scratch/entk/venv/local/lib/python2.7/site-packages/radical/entk/appman/appmanager.py", line 888, in _synch$
onizer
    task_update(msg, '%s-sync-to-cb' % self._sid, props.correlation_id, mq_channel)
  File "/home/vagrant/scratch/entk/venv/local/lib/python2.7/site-packages/radical/entk/appman/appmanager.py", line 641, in task_up
date
    completed_task.from_dict(msg['object'])
  File "/home/vagrant/scratch/entk/venv/local/lib/python2.7/site-packages/radical/entk/task/task.py", line 780, in from_dict
    actual_type=type(d['executable']))
TypeError: Expected (base) type(s) <type 'str'>, but got <type 'list'>.
```

[...]

PR #323, opened 2019-03-01
Compatibility/exec

andre-merzky:
This PR reroutes the `from_dict` type checks to the setter type checks for the executable, thus fixing #322.


andre-merzky:
PS.: this contains commits from master, too - as soon as master is merged back into devel, this PR will become much smaller.

mturilli:
@andre-merzky, the error in Codacy is a false positive. Could you reformat so to avoid it? I was unable to understand why Travis complains, all the tests seem to pass. Any idea?

andre-merzky:
I don't see the codacy output - when I follow the link, it tells me that the branch is not covered.  If you see the message, can you paste it please?

The travis tests seem to hang - Vivek says that this happens now and then, and he just reruns them.  I am not sure I agree with that approach, but that's for another time to discuss ;-)

andre-merzky:
Pfft, systems...  I see it now.

That part of code wasn't really touched by my PR - it will disappear once the devel branch is synced with master.

PR #326, opened 2019-03-14
GPU CUDA thread type

jdakka:
To enable `EnTK` to set `gpu_thread_type` = `CUDA` as defined by this example in RP: https://github.com/radical-collaboration/hyperspace/blob/master/microscope/rp_summit.py#L95


jdakka:
I tested with this branch and it works as expected. Should I close the pull request?

andre-merzky:
> I tested with this branch and it works as expected. Should I close the pull request?

@vivek-bala should decide on how/when to merge and release it I guess.

iparask:
I think the change should also be tested. This PR drops the coverage based on Codecov. Coach does not seem to work, because it does not analyze the `devel` branch. @vivek-bala, is this on purpose?

andre-merzky:
> This PR drops the coverage based on Codecov.

Don't do this number-hinting, please.  Codecov is meaningless in those cases.   All the diff does is add to documentation, and add a value check...

andre-merzky:
Add test

PR #328, opened 2019-03-21
Feature/expand workflow writer

iparask:
This PR is doing the following:
1. Extends the write workflow function by getting as input the filename the user might want to write the workflow, and a boolean value as to whether or not the workflow will be written to a file or not.
2. Updates the test for this function
3. Fixes a bug found by pylint by adding an import for RP states.
4. Small changes due to error messages from pylint and flake8


src/radical/entk/utils/prof_utils.py
```diff
@@ -1,17 +1,11 @@
...
-import json
+import traceback
 import radical.utils as ru
 
 from radical.entk.exceptions import *
-import traceback
-import socket
 from radical.entk import states as res
-
+from radical.pilot import states as rps
```
iparask: This import was missing and line 50 would have failed

mturilli:
What use case(s) this PR serves?

iparask:
It serves the ability to create integration and unit test for applications that use EnTK for their workflow execution.

`write_workflow` function allows to write a workflow as a list of dictionaries, which then makes assertion very easy for tests.

andre-merzky:
Note that RP has a similar mechanism which is triggered by setting `RADICAL_PILOT_RECORD_SESSION` (see [here](https://github.com/radical-cybertools/radical.pilot/blob/devel/src/radical/pilot/session.py#L235)).  Should we consider using similar mechanisms?  RP basically writes pilot and task descriptions to disk when they are passed to the API when this env variable is set.  The resulting files can then be used to replay a session.  I used this mostly to debug EnTK session BTW, as that allows me to capture the pilots and tasks requested by EnTK :-)

vivek-bala:
Looks good to me!

@andre-merzky Yea, I think there might be similarities in what we are doing in EnTK and RP. As to whether they should be argument or an environment variable, I generally prefer arguments/configurations to objects than environment variables. But I'll let you guys decide.

mturilli:
The problem I have with envs is documentation. Arguments/configs document themselves to some extent via automatic API generation or explicit defaults in the config files. Given the current RP documentation, how would I know about `RADICAL_PILOT_RECORD_SESSION`?

andre-merzky:
Documentation is required in both cases.  Docs are worse in RP than EnTK, but apart from that, the env is not documented as that feature was not yet exposed to users at all, but only used internally...

andre-merzky:
We merge now for the release, and keep RCT consistency for a later discussion (it can be kept in a backward compatible way).

andre-merzky:
@iparask , @vivek-bala : it looks like travis tries to apply the python coding guidelines to `.travis.yml`, and (unsurprisingly) fails?

vivek-bala:
Yup, I have pinged Giannis on the slack devel channel about it as well. @iparask let me know if you need any help.

iparask:
Codacy fails with:
```
Seems like the devel is not being analyzed. To review this pull request enable the branch in the project settings.
```
@vivek-bala, can you fix it or should I not pay attention? Travis passes all the tests!

vivek-bala:
@iparask I have already disabled Codacy, but I think it remains for the PRs that were already opened. Ignore them.

andre-merzky:
Thanks all - I think travis is now in the state we wanted it to be in...

PR #338, opened 2019-04-22
changes for bringing WFprocessor into Amgr process + comments from PR…

vivek-bala:
This PR brings the WFprocessor component into the same process as Amgr. Also address the comments from PR #310 . 

Note: the suspend/resume functionality needs to be carefully merged with this PR, but this is for later.


vivek-bala:
I have run some examples to test the changes. Will look into the remainings tests based on what travis says for this PR.

@andre-merzky and @iparask thanks for all your comments and questions on #310 . But since the code has changed, I have addressed them in this PR. I think I have addressed all the comments from #310 (as applicable). You don't have to compare the two PRs, but simply review this one. Apologies for the overhead of the two, but hopefully my comments in this PR are helpful.

@andre-merzky let me know if you are comfortable to close #310 in lieu of this PR.

vivek-bala:
Note that the comment additions, indentation consistency are mostly for the WFprocessor (major change). I'll replicate it for the Amgr as well.

src/radical/entk/appman/appmanager.py
```diff
@@ -270,9 +270,28 @@ def run(self):
 
             self._prof.prof('amgr run started', uid=self._uid)
 
+
+            # Submit resource request if not resource allocation done till now or
+            # resubmit a new one if the old one has completed
+            if self._resource_manager:
+                res_alloc_state = self._resource_manager.get_resource_allocation_state()
+                if (not res_alloc_state) or (res_alloc_state in self._resource_manager.get_completed_states()):
```
andre-merzky: Brackets around the logical expressions are not needed, and line is too long. This can be written as:
```
if not res_alloc_state or \
    res_alloc_state in self._resource_manager.get_completed_states():
```

src/radical/entk/appman/appmanager.py
```diff
@@ -270,9 +270,28 @@ def run(self):
...
+                    self._resource_manager._submit_resource_request()
+                    res_alloc_state = self._resource_manager.get_resource_allocation_state()
+                    if res_alloc_state in self._resource_manager.get_completed_states():
+                        raise EnTKError(msg="Cannot proceed. Resource allocation ended up in %s"%res_alloc_state)
+
+            else:
+
+                self._logger.exception('Cannot run without resource manager, please create and assign a resource manager')
+                raise EnTKError(text='Missing resource manager')
+
```
andre-merzky: Some lines are too long.

src/radical/entk/appman/appmanager.py
```diff
@@ -270,9 +270,28 @@ def run(self):
...
+                res_alloc_state = self._resource_manager.get_resource_allocation_state()
+                if (not res_alloc_state) or (res_alloc_state in self._resource_manager.get_completed_states()):
+
+                    self._logger.info('Starting resource request submission')
+                    self._prof.prof('init rreq submission', uid=self._uid)
+                    self._resource_manager._submit_resource_request()
+                    res_alloc_state = self._resource_manager.get_resource_allocation_state()
+                    if res_alloc_state in self._resource_manager.get_completed_states():
+                        raise EnTKError(msg="Cannot proceed. Resource allocation ended up in %s"%res_alloc_state)
+
```
andre-merzky: `%` needs spaces before and after.

src/radical/entk/appman/appmanager.py
andre-merzky: Why do you have `EntkError(text=...)` and `EntkError(msg=...)`?

src/radical/entk/appman/appmanager.py
```diff
@@ -270,9 +270,28 @@ def run(self):
...
+            else:
```
andre-merzky: I would suggest to avoid nesting like this:
```
if something:
  do a
  do b
  do c
  if something_else:
    do d
  ...
  do x
else:
  raise(you need something)
```

instead perform the requirement checks all as early as possible, and leave the structure of the actual semantics self contained:

```
# check if we can do stuff
if not something:
  raise(you need something)

# do stuff
do a
do b
do c
if something_else:
  do d
...
do x
```

src/radical/entk/appman/appmanager.py
```diff
@@ -270,9 +270,28 @@ def run(self):
...
+
+                self._logger.exception('Cannot run without resource manager, please create and assign a resource manager')
+                raise EnTKError(text='Missing resource manager')
```
andre-merzky: `logger.exception` only makes sense in a `except` clause, as only there you have an exception trace to log.  I think this should be a `logger.error`?

src/radical/entk/appman/appmanager.py
```diff
@@ -281,9 +300,10 @@ def run(self):
                     raise EnTKError("RabbitMQ setup failed")
```
andre-merzky: You never use `setup` again.  So, I suggest to either do:
```
self._mqs_setup = self._setup_mqs()
if not self._mqs_setup:
  raise...
```

or:

```
if not self._setup_mqs():
  raise...
```

But I would actually suggest to move that whole logic into the `self._setup_mqs()` method itself:

```
def _setup_mqs(self):
  if self.__setup_mqs_ok:
    return  # nothing to do
  
  try:
    do a
    do b
    ...
    self._setup_mqs_ok = True
  except ...:
    raise('RabbitMQ setup failed')
```

and then the code here simplifies to 

```
    self._setup_mqs()
```

Error handling should be localized to where the error can happen, not where the method is being called.

src/radical/entk/appman/appmanager.py
```diff
@@ -849,15 +727,14 @@ def pipeline_update(msg, reply_to, corr_id, mq_channel):
 
                     msg = json.loads(body)
 
-                    self._prof.prof('received obj with state %s for sync' %
-                                    msg['object']['state'], uid=msg['object']['uid'])
+                    self._prof.prof('received obj with state %s for sync' % msg['object']['state'], 
+                                    uid=msg['object']['uid'])
```
andre-merzky: The  first arg for `prof.prof` should be a short event name, like `obj_rcv`.  You can add a longer text in the `msg` field - but will you ever parse this msg in aalytics?  the profiler is not a logger ;-)

src/radical/entk/appman/appmanager.py
```diff
@@ -878,65 +754,17 @@ def pipeline_update(msg, reply_to, corr_id, mq_channel):
 
                     msg = json.loads(body)
 
-                    self._prof.prof('received obj with state %s for sync' %
-                                    msg['object']['state'], uid=msg['object']['uid'])
+                    self._prof.prof('received obj with state %s for sync' %msg['object']['state'],
+                                    uid=msg['object']['uid'])
```
andre-merzky: same

src/radical/entk/appman/wfprocessor.py
```diff
@@ -100,145 +110,173 @@ def _initialize_workflow(self):
 
         except Exception, ex:
             self._logger.exception(
-                'Fatal error while initializing workflow: %s' % ex)
+                'Fatal error when initializing workflow: %s')
             raise
```
andre-merzky: `%s` is unresolved, `ex` is not used anymore?

src/radical/entk/appman/wfprocessor.py
```diff
@@ -100,145 +110,173 @@ def _initialize_workflow(self):
...
+                                             profiler=self._prof,
+                                             logger=self._logger,
+                                             reporter=self._report)
+
+                        # Get the next stage of this pipeline to process
+                        exec_stage = pipe.stages[pipe.current_stage - 1]
+
+                        if not exec_stage.uid:
+                            # TODO: Move parent uid, name assignment to
+                            # assign_uid function
```
andre-merzky: Isn't the code below already doing what the `TODO` requests?

src/radical/entk/appman/wfprocessor.py
```diff
@@ -100,145 +110,173 @@ def _initialize_workflow(self):
...
+
+                                local_transition(obj=exec_stage,
+                                                    obj_type='Stage',
+                                                    new_state=states.SCHEDULING,
+                                                    profiler=self._prof,
+                                                    logger=self._logger,
+                                                    reporter=self._report)
+
+                            exec_tasks = exec_stage.tasks
+
```
andre-merzky: The above structure is:
```
if state in [INITIAL, SCHEDULED]:
  if state == INITIAL:
    transition_to_SCHEDULED
  do something
```

I think this would be clearer as:

```
if state == INITIAL:
  transition_to_SCHEDULED

if state == SCHEDULED:
  do something
```
This assumes that `transition_to_SCHEDULED` is atomic.  Is it?

src/radical/entk/appman/wfprocessor.py
```diff
@@ -100,145 +110,173 @@ def _initialize_workflow(self):
...
+                        for exec_task in exec_tasks:
+
+                            if (exec_task.state == states.INITIAL) or \
+                                ((exec_task.state == states.FAILED) and
+                                    (self._resubmit_failed)):
+
```
andre-merzky: Some brackets above are not needed and only make (human) parsing more difficult.  I would suggest:
```
if exec_task.state == states.INITIAL or \
    (exec_task.state == states.FAILED and self._resubmit_failed):
```

But also, is `_resubmit_failed` only ever true if the task is `FAILED` in the first place, and thus implies that state already?  If not, shouldn't that be the case?

src/radical/entk/appman/wfprocessor.py
```diff
@@ -100,145 +110,173 @@ def _initialize_workflow(self):
...
+                                # We store the stages since the stages the
+                                # above tasks belong to also need to be
+                                # updated. If its a task that failed, the
+                                # stage is already in the correct state
+                                if exec_task.state == states.FAILED:
+                                    continue
+                                if exec_stage not in scheduled_stages:
+                                    scheduled_stages.append(exec_stage)
+
```
andre-merzky: At this point you have:
```
def ...:
  try:
    while:
      for:
        with:
          for:
            if:
             if:
```
which is hard to keep in mind when reading this code.  One option would be to make the thread main routine short and self contained, like:

```
def _enqueue(self):
  try:
    while not terminated:
      work()
  except:
     oops
```

which keeps the except closer than a gazillion lines to the try ;-)  And also lets `work` focus on the actual semantics.

src/radical/entk/appman/wfprocessor.py
```diff
@@ -100,145 +110,173 @@ def _initialize_workflow(self):
...
-                    # Put the task on one of the pending_queues
-                    workload_as_dict = list()
+                    # Tasks of the workload need to be converted into a dict
+                    # as pika can send and receive only json/dict data
+                    workload_as_json = list()
                     for task in workload:
-                        workload_as_dict.append(task.to_dict())
-                    workload_as_dict = json.dumps(workload_as_dict)
+                        workload_as_json.append(task.to_dict())
+                    workload_as_json = json.dumps(workload_as_json)
```
andre-merzky: This could be written as 
```
  wl_json = [task.to_dict() for task in workload]
```
or even:
```
  wl_json = json.dumps([task.to_dict() for task in workload])
```
if you like that kind of thing...

src/radical/entk/appman/wfprocessor.py
```diff
@@ -249,328 +287,235 @@ def _enqueue(self, local_prof):
...
         except Exception, ex:
 
-            self._logger.exception('Error in enqueue-thread: %s' % ex)
+            self._logger.exception('Error in enqueue-thread')
             try:
                 mq_connection.close()
             except Exception as ex:
                 self._logger.warning('mq_connection not created, %s' % ex)
 
             raise
```
andre-merzky: If you catch an exception, log it as exception, and raise it again, then the same exception and the same trace get logged twice.

src/radical/entk/appman/wfprocessor.py
```diff
@@ -249,328 +287,235 @@ def _enqueue(self, local_prof):
...
             raise KeyboardInterrupt
 
         except Exception, ex:
 
-            self._logger.exception('Error in enqueue-thread: %s' % ex)
+            self._logger.exception('Error in enqueue-thread')
             try:
                 mq_connection.close()
             except Exception as ex:
                 self._logger.warning('mq_connection not created, %s' % ex)
```
andre-merzky: `not closed` I bvelieve?

src/radical/entk/appman/wfprocessor.py
```diff
@@ -249,328 +287,235 @@ def _enqueue(self, local_prof):
...
-                    self._dequeue_thread_terminate.set()
-                    self._dequeue_thread.join()
-
-            self._logger.info('WFprocessor process terminated')
-
-            self._logger.exception('%s failed with %s'%(self._uid, ex))
-            raise EnTKError(ex)
+            self._logger.exception('Execution failed in post_exec \
+                                    of stage %s' % stage.uid)
+            raise
```
andre-merzky: same

src/radical/entk/appman/wfprocessor.py
```diff
@@ -249,328 +287,235 @@ def _enqueue(self, local_prof):
...
         try:
 
-            local_prof = ru.Profiler(
-                name='radical.entk.%s' % self._uid + '-proc', path=self._path)
-
-            local_prof.prof('wfp process started', uid=self._uid)
-
-            self._logger.info('WFprocessor started')
+            self._logger.info('Executing post-exec for stage %s' % stage.uid)
+            self._prof.prof('Adap: executing post-exec', uid=self._uid)
```
andre-merzky: I would suggest event names like `post_exec_start` and `post_exec_stop`.

src/radical/entk/appman/wfprocessor.py
```diff
@@ -249,328 +287,235 @@ def _enqueue(self, local_prof):
...
+            self._logger.exception('Error in dequeue-thread')
             try:
                 mq_connection.close()
-            except:
-                self._logger.warning('mq_connection not created')
+            except Exception as ex:
+                self._logger.warning('mq_connection not created, %s' % ex)
 
-            raise EnTKError(ex)
+            raise
```
andre-merzky: same

src/radical/entk/appman/wfprocessor.py
```diff
@@ -579,35 +524,36 @@ def start_processor(self):
...
-                self.terminate_processor()
-                raise
+        except Exception, ex:
 
-        else:
-            self._logger.warn(
-                'Wfp process already running, attempted to restart!')
+            self._logger.exception('WFprocessor failed')
+            self.terminate_processor()
+            raise
```
andre-merzky: same

src/radical/entk/utils/init_transition.py
```diff
@@ -39,3 +39,22 @@ def transition(obj, obj_type, new_state, channel, queue, profiler, logger):
...
+        msg = None
+
+    obj.state = new_state
+    profiler.prof(  'advance',
+                    uid=obj.uid,
+                    state=obj.state,
+                    msg=msg)
+    reporter.ok('Update: ')
+    reporter.info('%s state: %s\n' %(obj.luid, obj.state))
+    logger.info('Transition of %s to new state %s successful' % (obj.uid, new_state))
```
andre-merzky: Those reporter messages are very noisy for large numbers of tasks.  But this is unrelated to the PR semantics, just saying :-)

mturilli:
Just a reminder that we agreed to merged this after we will have released the pause/restart feature.

src/radical/entk/appman/appmanager.py
vivek-bala: Thanks Andre. I've done this extensively for the wfprocessor and brought down the indentation levels. I'm doing it for the amgr as well, not complete yet.

src/radical/entk/appman/appmanager.py
vivek-bala: Fair point. Will do :+1:

vivek-bala:
Thanks for the quick feedback Andre! Going through them..

vivek-bala:
@mturilli @andre-merzky Sure, but the pause/resume functionality will change a bit with this PR. I agree this PR should not go out for the next release. I will let you both decide about the PR with pause/resume.

src/radical/entk/appman/wfprocessor.py
vivek-bala: Well, I want to move the parent uid assignment to the _assign_uid function as well.
Currently, the _assign_uid function only assigns parent uids to contents (tasks in this 
case) of the object (stage), not the object itself.

src/radical/entk/appman/wfprocessor.py
vivek-bala: Cool, will do. Yes, the assumption is correct.

vivek-bala: Well, the following is the equivalent:

```
if state == INITIAL:
    transition_to_SCHEDULED
    do something

if state == SCHEDULED:
    do something
```

Would you prefer that? I personally prefer as it exists now, but happy to go with something that works for you.

src/radical/entk/appman/wfprocessor.py
vivek-bala: _resubmit_failed is a flag that the user sets. So if the flag is set and the task state is FAILED, we have to resubmit the task. Does that answer your question?

andre-merzky:
> @mturilli @andre-merzky Sure, but the pause/resume functionality will change a bit with this PR. I agree this PR should not go out for the next release. I will let you both decide about the PR with pause/resume.

I would propose to release suspend/resume ASAP, to get it off the table, and to avoid further diverging branches.  It is more urgent than the threadification also.  The merge needs to be done anyway, it does not matter much in what direction we have to resolve the conflicts.

My $0.02, Andre.

src/radical/entk/appman/appmanager.py
andre-merzky: Yep, I saw that! :-D

src/radical/entk/appman/wfprocessor.py
vivek-bala: Well, that reduced one level of indentation. As long as the code is within the margins, it works for me.
But seeing that you would prefer them shorter, maybe we can do the following:

```
def _enqueue(self):
  try:
    while not terminated:
      workload = create_workload()
      if workload:
        execute_workload()
  except:
     oops
```

There is a logical separation in what work() does as well and I would find it more meaningful. What do you think?

src/radical/entk/appman/wfprocessor.py
vivek-bala: Yup, makes sense :+1:

src/radical/entk/appman/wfprocessor.py
vivek-bala: logging the exception doesn't the parent object that an exception was encountered though right? I raise it again so that parent object essentially stops and does not try to continue. Maybe there is another/better way to do that?

src/radical/entk/utils/init_transition.py
vivek-bala: True. At that point we can ask users to disable the reporting or have multiple verbosity for reporting as we do for log messages though, right? For now, I am keeping it consistent with Matteo implemented.

src/radical/entk/appman/appmanager.py
vivek-bala: I will check again, but there is only one point in the WFProcessor where I did not want to check for the negation because I think that logic will be harder to process - this is when create a workload of ready tasks or failed tasks.

If there is any other point where you saw scope for this, can you point it please? :)

src/radical/entk/appman/wfprocessor.py
andre-merzky: Yes, that works for sure!  Don't take all my comments at face value BTW, I know it's not always possible / sensible to adhere to the guidelines religiously - but your suggestion would make the code more readable IMHO!

vivek-bala:
> > @mturilli @andre-merzky Sure, but the pause/resume functionality will change a bit with this PR. I agree this PR should not go out for the next release. I will let you both decide about the PR with pause/resume.
> 
> I would propose to release suspend/resume ASAP, to get it off the table, and to avoid further diverging branches. It is more urgent than the threadification also. The merge needs to be done anyway, it does not matter much in what direction we have to resolve the conflicts.
> 
> My $0.02, Andre.

Sure makes sense. You're right, the merge will be complicated either way. It was more about whether you want to release suspend/resume in the next cycle. I am okay going with your suggestion as you're making the next release.

Just a suggestion: My hesitation is that there has only been one user that has tested that functionality (apart from you and I), I wanted to see if we can give it some more time in testing.The branches will diverge, but I feel like it would in our control -- but you have more experience than me with this.  What do you think?

src/radical/entk/appman/wfprocessor.py
vivek-bala: Cool, will make the change then :+1:

src/radical/entk/appman/appmanager.py
andre-merzky: Only here in the app-manager I think.

src/radical/entk/appman/wfprocessor.py
andre-merzky: then I don't understand the transition.  I would expect the state to be `SCHEDULED` after, well,  transition to `SCHEDULED`, and then the second `if` would apply and do the something?

```python
if state == INITIAL:
    transition_to_SCHEDULED
    # --> state is now SCHEDULED, and the 'if' below will apply

if state == SCHEDULED:
    do something
```

src/radical/entk/appman/wfprocessor.py
andre-merzky: yes

src/radical/entk/appman/wfprocessor.py
andre-merzky: I am sorry, I don't understand your comment :(

> logging the exception doesn't the parent object that an exception was encountered though right?

Is there a verb missing?  What is expected to (not) happen to the parent object?  What do you mean with parent object - parent of what?

>  I raise it again so that parent object essentially stops and does not try to continue. 

An object does not continue - do you mean this method?  That will sure be aborted on `raise` - but it will also be aborted if you let the exception fall through.

src/radical/entk/utils/init_transition.py
andre-merzky: One could consider a similar mechanism as RP, where state information are less verbose (single character like `+` or `-`).  Or one only reports pipelines and stages, not tasks?

Have a look at the `async_repex` example in `fix/issue_296`: it `reports` pipelines and tasks to `dump.log`.  Do you want me to try to extend that to stages, too, and to pipe that through the reporter?

src/radical/entk/appman/appmanager.py
vivek-bala: fixed

src/radical/entk/appman/appmanager.py
vivek-bala: fixed

src/radical/entk/appman/appmanager.py
vivek-bala: fixed

src/radical/entk/appman/appmanager.py
vivek-bala: this was an error, it should be 'msg'. Fixed

src/radical/entk/appman/appmanager.py
vivek-bala: I have used this in all places (but one as mentioned above). The above lines are required to make sure to not proceed/exit if the pilot reaches FAILED (or DONE incorrectly).

src/radical/entk/appman/appmanager.py
vivek-bala: That's correct. Thanks. Fixed.

src/radical/entk/appman/appmanager.py
vivek-bala: Tbh I was doing that for some of the adaptation events :P But I shouldn't. Will change now.

src/radical/entk/appman/wfprocessor.py
vivek-bala: No longer exists.

src/radical/entk/appman/wfprocessor.py
vivek-bala: Yes, in the new code, I have implemented it in that manner. Fixed.

src/radical/entk/appman/wfprocessor.py
vivek-bala: Actually, it exists and is fixed.

src/radical/entk/appman/wfprocessor.py
vivek-bala: Uhmm, I don't understand myself either :-| Think I missed several words, let me try again!

Function X calls function Y. Function Y raises an exception which is logged. But upon that exception, I want function X to stop executing as well, so I raise it again.

```python
def func_x():
    func_y()
    print 'not to print'

def func_y():
    try:
        x = 1/0
    except:
        print 'bad math'

>>> func_x()
bad math
not to print
```

vs

```python
def func_x():
    func_y()
    print 'not to print'

def func_y():
    try:
        x = 1/0
    except:
        print 'bad math'
        raise

>>> func_x()
bad math
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "<stdin>", line 2, in func_x
  File "<stdin>", line 3, in func_y
ZeroDivisionError: integer division or modulo by zero
```

Does that help? If you meant, we shouldn't be logging again in function X (although logging is not there in my example), I may agree, but I need to check.

src/radical/entk/appman/wfprocessor.py
vivek-bala: Done

src/radical/entk/utils/init_transition.py
vivek-bala: I would be in favor of reporting just pipelines and stages, but at a different reporter level (can we have reporter levels 1,2,3 or so?).

But also, I would like to leave it as is in this PR and revisit this later. Would that be work?

src/radical/entk/appman/wfprocessor.py
andre-merzky: It does.  My suggestion is to just not except it:
```
>>> def func_x():
...     func_y()
...     print 'not to print'
...
>>> def func_y():
...     x = 1/0
...
>>> func_x()
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "<stdin>", line 2, in func_x
  File "<stdin>", line 2, in func_y
ZeroDivisionError: integer division or modulo by zero
```

Whover calls `func_y()` / `_execute_post_exec()` needs to be aware that it can raise an exception, needs to except that, and will lock the **same** stacktrace again.  I don't see the benefit of the double try/catch?

andre-merzky:
I tried to run the test suite the same way as done by Travis.  It succeeded once (running about 2 hours though), but the second time it hangs on test_issue_26.  The only error I found in the logs is:
```
2019-05-30 13:48:01,155: radical.entk.task_processor: MainProcess                     : MainThread     : ERROR   : Failed to get input list of files from task, error: Expected (base) type(s) <class 'radical.entk.task.task.Task'>, but got <type 'list'>.
```
(repeated multiple times).  An hour later is now seems unlikely that the tests will continue.

@vivek-bala : any suggestion?

andre-merzky:
I changed the test so it's easy to run it stand-alone.  The same test does succeed in master, so there seems to be a regression introduced. :-(

andre-merzky:
I tried to track this down to a specific commit, but failed to do so.  But from trying to understand the code, my assumption is that the following is happening:

The creation of the app manager does not create the wfproc threads - that happens when `run` is called.    But when `run` is called again, those components are created *again*, and the old ones are abandoned.  Then on termination, those orphaned threads have no means of getting joined, and all the main thread will hang.

I assume that the same actually happened when the wfproc was process based - but then the missing `join` (or rather `wait`) on the orphaned process will make it a zombie, and it not hang up the parent process.

Why is the process/thread setup not done during class creation?  Are there any information missing at that point?  I would not think so, as in the test code, the only information being added is the workflow, which should have no influence on the thread creation.

@vivek-bala , please ping back if you want to look at this, or if I should dig further.

andre-merzky:
The tests fail only for travis resource issues now, I consider this PR ready to be merged.

mturilli:
Seeing that the test is 'useless' and 'stupid' is there a way to do better instead of simply ignoring or deleting it?

andre-merzky:
> Seeing that the test is 'useless' and 'stupid' is there a way to do better instead of simply ignoring or deleting it?

Sorry, this is just me being annoyed.  I did not want to open the can of worms to discuss those tests in the context of this PR (or actually not in general right now), but I am getting somewhat frustrated in fixing tests which don't test useful things.

Point in case: the constructor sets `self.abc = True`, the test checks `assert (self.abc == True)`.  What this basically checks is if Python assignments work - which is not a useful test.  Since the tests in EnTK take ages to finish, it is always annoying to have to come back after an hour to fix something like that, and then push and wait for another hour only to be told there is another test like that which I missed... :-P  Context switching is difficult, and there are too many concurrent things already...

Anyway, thanks for listening :-)  This will all improve once we separate integration and unit tests, and I can actually (a) run tests in reasonable time, and (b) run tests locally.

src/radical/entk/appman/wfprocessor.py
```diff
@@ -86,6 +86,9 @@ def _advance(self, obj, obj_type, new_state):
         transition `obj` of type `obj_type` into state `new_state`
         '''
 
+        # NOTE: this is a local operation, no queue communication is involved
+        #       (other than the `_advance()` in the TaskManager classes, which
+
```
mturilli: Is this truncated?

tests/test_component/test_tmgr_advance.py
```diff
@@ -1,50 +1,60 @@
...
+
 import pika
-from radical.entk import Task, Stage, Pipeline
-import radical.utils as ru
 import os
-from threading import Thread
 import json
+
+import threading as mt
+
```
mturilli: :D

mturilli:
Not a problem Andre, really! Frustration is perfectly legit and thank you for giving more details. I would transform your comment into a ticket for the next release and then I would move on and merge.

andre-merzky:
> Not a problem Andre, really! Frustration is perfectly legit and thank you for giving more details.

Thanks Matteo! :-)

>  I would transform your comment into a ticket for the next release and then I would move on and merge.

I began to add `FIXME` comments in some places in the code and tests, where I think we need to test better or where the purpose of tests is not clear.  But no matter my frustration, the tests for the other RCT  layers are more urgent.  

Either way though, I am willing to spend some more days on trying to get the tests / code fixed so that tests actually pass.

andre-merzky:
@AymenFJA @lee212 @iparask : the EnTK tests are as good as they get for now.  4 tests are still failing, but those seem unrelated to code changes, but rather related to test setup.  I am going to merge this PR now, since the release needs to go out and this is the main change for it.  Can you guys please take care of the remaining test problems?  Thanks!

Issue #330, opened 2019-03-29
change amgr terminate method


vivek-bala:
amgr.resource_terminate() -> amgr.terminate()

since we terminate all components, not just the resource

iparask:
Introduce a private method called `_terminate`, a public method with the same name. The private method should be called by `resource_terminate` and introduce a deprecation method

iparask:
Based on our last discussion, `_terminate` should be merged with the public method.

PR #342, opened 2019-05-07
removed `_terminate` based on #330, pylint and flake8

andre-merzky:
@iparask : why does the `remove _terminate` branch rename the `write_workflow` method?    Also, the commit seems fairly large for the method remove.  Am I missing something?


iparask:
Hello Andre, this branch was not ready for a PR yet. :)

While I was running pylint for the `appmanager`, I got a warning saying:
```W: 48,17: Redefining name ‘write_workflow’ from outer scope (line 11) (redefined-outer-name)```

As I was trying to understand why this was happening, I found that `write_workflow` was used as a imported function name and a constructor parameter.

I discussed the issue with @vivek-bala and based on his recommendation, I changed the function name. In addition, `write_workflow` used only in the appmanager, so the impact from the name changing is minimal.

Does that help?

**Comment by andre-merzky** [comment]
Thanks Ioannis, that helps.  Yeah, I was aware that it's not ready for merge yet, no problem.

FWIW, another way to remove name conflicts like that is, instead of
```
from radical.entk.utils.prof_utils import write_workflow
write_workflow(...)
```

to use 
```
import radical.entk.utils as reu
reu.prof.write_workflow(...)
```

which gives the code reader some context of what is being called (versus needing to resolve the difference between `write_workflow` and `writing_workflow` in some way).

**Comment by iparask** [comment]
I made the change as recommended,

File: src/radical/entk/appman/appmanager.py [hunk 1]
```diff
@@ -384,14 +386,12 @@ def run(self):
...
-                    in the appmanager and start the processor.
-                    """
+                    # If WFP dies, both child threads are also cleaned out.
+                    # We simply recreate the wfp object with a copy of the workflow
+                    # in the appmanager and start the processor.
 
                     self._prof.prof('recreating wfp obj', uid=self._uid)
-                    self._wfp = WFProcessor(
+                    self._wfp = WFprocessor(
                         sid=self._sid,
```
andre-merzky: How has that ever worked?  The same code is in `master`.  I am puzzled...

andre-merzky: @vivek-bala : I really could not figure this one out - this seems to have been typed like that since, like, forever.  Any idea why python did not complain?

iparask: I think the code was never reached from any use case, and as a result python's interpreter never tried to run it.

andre-merzky:
This PR is getting out of sync, and specifically has the danger to conflict with the WFProc changes.  Please review ASAP.

src/radical/entk/appman/appmanager.py
vivek-bala: Giannis is correct. I think that code part was not reached in most occasions.

PR #344, opened 2019-05-07
Fix/exec scalar

andre-merzky:
This changes task executable specification from list (which is semantically wrong) to string.  The change is backward compatible.


andre-merzky:
This PR has a large number of trivial changes, and thus has a high probability of conflicting with other PRs.   Please review ASAP.

src/radical/entk/task/task.py
```diff
@@ -488,12 +488,18 @@ def pre_exec(self, value):
...
     @executable.setter
     def executable(self, value):
-        if isinstance(value, list):
-            self._executable = value[0]
-        elif isinstance(value, str):
+
+        elif isinstance(value, list):
+            value = value[0]
+
+        if value is None:
```
vivek-bala: Why would value be None? This means some user wrote ```task.executable = None```. I think this should raise the TypeError, we can't really execute a task without an executable.

andre-merzky: It can become `None` from `value[0]` above, thus the check.  But you are right, the two lines should go away and the type check should kick in on `None`.  I'll fix that.

PR #345, opened 2019-05-07
Fix/write workflows

andre-merzky:
The current implementation of `write_workflow` only writes the last workflow.  This fix ensures that all submitted workflows are recorded for a session.


src/radical/entk/utils/prof_utils.py
```diff
@@ -219,42 +220,53 @@ def get_session_description(sid, src=None):
     return desc
 
 
-def write_workflow(workflow, uid, workflow_fout='entk_workflow', fwrite=True):
+# ------------------------------------------------------------------------------
+#
+def write_workflow(workflows, uid):
+def write_workflow(workflows, uid, workflow_fout='entk_workflow', fwrite=True):
```
iparask: @andre-merzky did you happen to define twice the same function?

andre-merzky: Thanks - fixed!1

andre-merzky:
@iparask : I'm afraid this needs a new review, will go into the next release.

andre-merzky:
@iparask : This should be ready for a new review.

PR #346, opened 2019-05-07
Travis YML update

iparask:
Travis YML file is updated to include only `py` files for flak8 and pylint and also tries to disable them if there is no diff in a `py` file.

This is done on whether `$DIFF` is empty or not. We should not do this on `$LOG` because `$LOG` can be non empty if we change a non python file as in this PR.


iparask:
Done the suggested changes!

andre-merzky:
Thanks!

PR #348, opened 2019-05-22
Update .travis.yml

iparask:
Correcting quotes and zero test in travis YML


andre-merzky:
Oh, wait - you will need to use double quotes to get `$DIFF` to expand...

PR #354, opened 2019-06-05
attempt to fix wfp creation

andre-merzky:
This is an attempt to fix the WFP creation, and ensure it is only created once.  

Test issue_26 now terminates, but still fails because of some uid issues which I assume are unrelated (but I don't know).  Anyway, @vivek-bala , please have a look at this PR if that makes any sense.

The apparently WFP can't be started in `__init__` because, well, reasons: at that point no resource is assigned which means that no resource manager exists which means that the task manager cannot be created.  I think those instantiations need to be disentangled - the WFP should have nothing to do with the task manager, it can begin working even if there is no task manager or resource listening on the other end of the queue?

### DIFF
--- a/src/radical/entk/appman/appmanager.py
+++ b/src/radical/entk/appman/appmanager.py
@@ -107,6 +107,10 @@ class AppManager(object):
         self._shared_data = list()
         self._wfp = None
         self._sync_thread = None
+        self._terminate_sync = Event()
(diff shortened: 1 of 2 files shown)


vivek-bala:
Hey Andre, I'll take a look at this later today. I'll keep you posted.

PR #358, opened 2019-06-23
Feature/from dict

andre-merzky:
This PR adds the ability to initialize a task from a dictionary (similar to RP where we can initialize a unit from a dict).  The change is backward compatible.  I need it (meaning it makes my life easier) in repex to red task descriptions from json files.

The change is actually trivial, and all semantic changes are in the first commit.  The second commit is all about coding guidelines in the two files touched by the first commit, but also includes fixes and improvements to the type checks of the class attributes.


andre-merzky:
### pyflake:

> /test_component/test_task.py:71:8: E741 ambiguous variable name 'l'

This is a stupid one, can we remove this?  This is for people who use fonts where `l` looks like `1`, `L` looks like `1` etc.  I don't think we should work around font choices...

I fixed the other pyflake error, missed that when walking through the test.

### pylint:

This complains about access to private variables, but I think that this should be allowed in tests.  Can we disable this check for all tests?

The other complaint is wrong, too - it seems to misunderstand what the decorator above the method definition does.  Not sure how to fix.

### pytest

As for the pytest problem: this is the same old resource problem on travis, I have no idea what to do about it.

iparask:
> ### pyflake:
> > /test_component/test_task.py:71:8: E741 ambiguous variable name 'l'
> 
> This is a stupid one, can we remove this? This is for people who use fonts where `l` looks like `1`, `L` looks like `1` etc. I don't think we should work around font choices...
> 
> I fixed the other pyflake error, missed that when walking through the test.
> 

I'm of the opinion that single letter variables apart from when they are used as indices are very difficult to follow. It says ambiquous variable name "L" in lower case. I looked at the code and it is not easy to understand what "l" represents. I think avoiding single letter variable names will help with maintaining the code in an easier fashion.

> ### pylint:
> This complains about access to private variables, but I think that this should be allowed in tests. Can we disable this check for all tests?
> 
> The other complaint is wrong, too - it seems to misunderstand what the decorator above the method definition does. Not sure how to fix.

Yes we can fix that in the test. At the beginning of the test file, you can add :
``` python
# pylint: disable=protected-access, unused-argument
```

and the linter will not complain for these two in the tests.

andre-merzky:
> I'm of the opinion that single letter variables apart from when they are used as indices are very difficult to follow. [...] I looked at the code and it is not easy to understand what "l" represents. I think avoiding single letter variable names will help with maintaining the code in an easier fashion.

Really?  This is how its used:

```
@given(s=st.text(),
       l=st.lists(st.text()),
       i=st.integers().filter(lambda x: type(x) == int),
       b=st.booleans())
def method(s, l, i, b):

       data_type = [s, l, i, b]
      for data in data_type:
```

  Come on, it written there: string `s`,  list `l`, integer `i`, boolean `b`, and placeholder `x`.  What would be a suitable replacement which is similarly memnonic and short?  Is that really something worthwhile to spend time on thinking about?  Is this better:

```
@given(foo=st.text(),
       bar=st.lists(st.text()),
       baz=st.integers().filter(lambda pip: type(pip) == int),
       qux=st.booleans())
def method(foo, bar, baz, qux):

       data_type = [foo, bar, baz, qux]
      for data in data_type:
```


>  It says ambiguous variable name "L" in lower case.

Sorry, I mean `I`.  I looked into the definition of E.741 because I did not understand what it complains about.  Its not complaining about single letters, its complaining about typesetting of `l` `I`and `1` being too similar, so about font choice.  And yes, for some reason it complains about uppercase `I` too.  And also `1` is an invalid variable name.  I think its braindead, really.

andre-merzky:
I checked the tests again.  These two are failing: `tests/test_component/test_tmgr_2.py::test_tmgr_rp_tmgr` and `tests/test_component/test_tproc_rp_2.py::test_create_task_from_cu`.  

`test_tmgr_rp_tmgr`: this also fails for me locally, but also fails in master.  It fails with a different error than in travis though: locally it complains about a queue in RabbitMQ not being found.  I do not know how to address this, but also don't think that this is related to the PR `from_dict()` feature, as that code path is not invoked.

`test_create_task_from_cu` works locally and fails on travis, presumably due to resource issues (a pty channel does not come up).

iparask:
I see. The tests need to be fixed, but since they are not relevant with the PR then we are okay. Also silence the error for `I` if you want.

andre-merzky:
> The tests need to be fixed

I don't know how.  The tests works even when using ssh access to the travis env - just not in automated mode  (the first test still prints the rabbitmq error, but it actually finishes with success).

Any advise?  Thanks.

iparask:
None what so ever... I think it might something to do with how ssh and internet is configured for non debug sessions. This says even more that unit tests need to be separated from integration tests.

andre-merzky:
@mturilli , @iparask : should we disable the failing tests?  If the test suite regularly does not pass in travis, then it is not useful, really, and we will get used to merging with a failing test.  So, should we bite the bullet and move integration tests into a separate subtree, so that we can trigger them only when so wanted?

andre-merzky:
@iparask , @mturilli : ping

PR #361, opened 2019-07-13
battle test resource issues

andre-merzky:
Run some tests in different processes to reduce resource consumption per process.  This specifically gets the integration tests going again, but also speeds up overall runtime.


andre-merzky:
This is not ready for merge, as it now competes on RP VE creation.  This needs resolving in RP.

PR #363, opened 2019-07-18
fix test setup

andre-merzky:
fix dburl and var name

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,7 +6,6 @@ os:
 
 env:
   global:
-    - RADICAL_PILOT_DBURL="mongodb://entk:entk123@127.0.0.1:27017/entk_test"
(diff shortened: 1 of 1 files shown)


iparask:
@mturilli, @andre-merzky do you think it makes to silence some of the integration tests? They are producing errors and I'm not sure it helps to having them run since the errors are not related to EkTK.

andre-merzky:
I am not sure what to do about this.  I think the set of failing tests is somewhat random, since they seem to be caused by resource problems - so silencing one set will likely just mean that others fail.  And also I don't think we should get into the habit of silencing tests which fail.  I guess it comes down to the discussion we had: travis should be used for unit testing, not integration testing.

I think some of these problems are fixed in #361 - but you see the integration test there failing as well, for a reason which I do not understand at all (failure to remove ~/.netrc, which is a file we do not explicitly touch in our stack, ever...)

PR #365, opened 2019-08-15
Fixed mistake using code-block

wilkinson:
This isn't really sophisticated or anything. I'm just reading through the documentation, and there was a minor mistake in markup.


andre-merzky:
Thank you!  Fixes like this are very welcome :-)

PR #367, opened 2019-09-04
Updated docs because "entk-version" command no longer exists

wilkinson:
Hello

This important part of this pull request updates the docs (same "Installation" page again) to reflect that the "entk-version" command has been replaced by the "radical-entk-version" command. This is important to update because it's actually the command recommended to check to make sure you installed everything correctly.

Also, I fixed my name and email on a previous commit (I think...).


andre-merzky:
Thank you! :-)

PR #378, opened 2019-10-08
Replaced RabbitMQ hostname,port with URL

wilkinson:
I changed the interfaces not to use hostname and port for RabbitMQ but instead to use a connection string URL. This will break other programs that used to specify hostname and port. Those programs should still be able to run if they use a connection string that explicitly writes out the assumed default parameters, along the lines of "amqp://guest:guest@localhost:5672/%2F" (but I haven't tested that, either).

### DIFF
--- a/src/radical/entk/appman/appmanager.py
+++ b/src/radical/entk/appman/appmanager.py
@@ -37,8 +37,6 @@ class AppManager(object):
 
     :Arguments:
         :config_path:     Url to config path to be read for AppManager
-        :hostname:        host rabbitmq server is running
-        :port:            port at which rabbitmq can be accessed
         :reattempts:      number of attempts to re-invoke any failed EnTK
                           components
         :resubmit_failed: resubmit failed tasks (True/False)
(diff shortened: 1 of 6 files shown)


wilkinson:
I definitely forgot to update tests in the TravisCI config file. This is definitely going to fail tests.

mturilli:
@wilkinson thank you for the patch! Should we keep `pika.ConnectionParameters` and use the `credentials` parameter? This would be back compatible and, most importantly, would not change the current API but just extend it. This is important because EnTK is used in production by multiple use cases already. 

From https://pika.readthedocs.io/en/stable/modules/credentials.html:

> To authenticate with Pika, create a PlainCredentials object passing in the username and password and pass it as the credentials argument value to the ConnectionParameters object.

wilkinson:
@mturilli By all means, please ignore this pull request if it's troublesome. I'm not emotionally attached to it. I have no intention of altering the API!

The reason I didn't keep `pika.ConnectionParameters` was because I didn't want to handle parsing the URL string myself. Parsing is easy, but it sometimes bites me -- kind of like how RADICAL-Pilot has some troubles with query parameters and MongoDB right now.

Your suggestion sounds better, and certainly it is more responsible. I don't know if it's better to kill this pull request and make another, or else to pile more commits onto it. Let me know.

wilkinson:
I'm reading the documentation now, and I see what you mean. I was in a hurry and hadn't planned to submit a pull request for the previous code until Andre encouraged it. I much prefer the approach you have described. Let me know how to proceed with the Git history, and I can re-do it with the correct objects.

wilkinson:
I'll kill this for now, and I'll plan to send a different pull request instead.

andre-merzky:
Sorry for the late reply!

Quick note that you don't need to parse the URL, we have a helper class for that:
https://github.com/radical-cybertools/radical.utils/blob/devel/src/radical/utils/url.py  :-)

PR #379, opened 2019-10-10
Accept username/password credentials for RabbitMQ

wilkinson:
So, this pull request is better than the last one, but it may still have some problems. It extends the `AppManager` constructor to accept `rmq_url` as an input parameter. The original `hostname` and `port` parameters have been preserved, and they will override `rmq_url` if specified at the same time. To do this, a `pika.connection.URLParameters` object is created and then its values are overridden. Then, instead of passing the `hostname` and `port` from function to function in various forms, the connection parameters object, named `rmq_conn_params`, is passed instead. 

This makes a lot of things a lot easier, but the downside is that things might get weird if you ever try to create the downstream objects directly, such as `WFProcessor`. This may cause the tests to fail, if you have unit tests that test the previous interfaces. It was necessary to change the interfaces downstream from `AppManager`, however, or else credentials would not be propagated.

### DIFF
--- a/src/radical/entk/appman/config.json
+++ b/src/radical/entk/appman/config.json
@@ -1,5 +1,7 @@
 
 {
+    "hostname"        : "localhost",
+    "port"            : 5672,
     "reattempts"      : 3,
     "resubmit_failed" : false,
     "autoterminate"   : true,
@@ -9,7 +11,6 @@
                          "db_cleanup"      : false },
     "pending_qs"      : 1,
     "completed_qs"    : 1,
-    "rmq_cleanup"     : true,
-    "rmq_url"         : "amqp://guest:guest@localhost:5672/%2F"
+    "rmq_cleanup"     : true
 }
 
--- a/src/radical/entk/appman/config.json
+++ b/src/radical/entk/appman/config.json
@@ -10,6 +10,9 @@
     "pending_qs"      : 1,
     "completed_qs"    : 1,
     "rmq_cleanup"     : true,
-    "rmq_url"         : "amqp://guest:guest@localhost:5672/%2F"
+    "hostname"        : "localhost",
+    "port"            : 5672,
+    "username"        : "guest",
+    "password"        : "guest"
 }
 
--- a/src/radical/entk/appman/config.json
+++ b/src/radical/entk/appman/config.json
@@ -1,5 +1,9 @@
 
 {
+    "hostname"        : "localhost",
+    "port"            : 5672,
+    "username"        : "guest",
+    "password"        : "guest",
     "reattempts"      : 3,
     "resubmit_failed" : false,
     "autoterminate"   : true,
@@ -9,10 +13,6 @@
                          "db_cleanup"      : false },
     "pending_qs"      : 1,
     "completed_qs"    : 1,
-    "rmq_cleanup"     : true,
-    "hostname"        : "localhost",
-    "port"            : 5672,
-    "username"        : "guest",
-    "password"        : "guest"
+    "rmq_cleanup"     : true
 }
 
--- a/src/radical/entk/appman/config.json
+++ b/src/radical/entk/appman/config.json
@@ -1,7 +1,5 @@
 
 {
-    "hostname"        : "localhost",
-    "port"            : 5672,
     "reattempts"      : 3,
     "resubmit_failed" : false,
     "autoterminate"   : true,
@@ -11,6 +9,7 @@
                          "db_cleanup"      : false },
     "pending_qs"      : 1,
     "completed_qs"    : 1,
-    "rmq_cleanup"     : true
+    "rmq_cleanup"     : true,
+    "rmq_url"         : "amqp://guest:guest@localhost:5672/%2F"
 }
 
--- a/src/radical/entk/appman/appmanager.py
+++ b/src/radical/entk/appman/appmanager.py
@@ -39,6 +39,8 @@ class AppManager(object):
         :config_path:     Url to config path to be read for AppManager
         :hostname:        host rabbitmq server is running
         :port:            port at which rabbitmq can be accessed
+        :username:        username to log in to RabbitMQ
+        :password:        password to log in to RabbitMQ
         :reattempts:      number of attempts to re-invoke any failed EnTK
                           components
         :resubmit_failed: resubmit failed tasks (True/False)
@@ -55,8 +57,6 @@ class AppManager(object):
                           True/False} when RTS is RP
         :name:            Name of the Application. It should be unique between
                           executions. (default is randomly assigned)
-        :rmq_url:         URL string to connect to RabbitMQ; its values will
-                          be overridden by :hostname: or :port:
     '''
 
     # --------------------------------------------------------------------------
@@ -65,6 +65,8 @@ class AppManager(object):
                  config_path=None,
                  hostname=None,
                  port=None,
+                 username=None,
+                 password=None,
                  reattempts=None,
                  resubmit_failed=None,
                  autoterminate=None,
@@ -72,8 +74,7 @@ class AppManager(object):
                  rts=None,
                  rmq_cleanup=None,
                  rts_config=None,
-                 name=None,
-                 rmq_url=None):
+                 name=None):
 
         # Create a session for each EnTK script execution
         if name:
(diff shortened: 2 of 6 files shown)


wilkinson:
Also, the advantage to carrying around the `pika.connection.URLParameters` object as opposed to just some `credentials` is that all the other server settings will propagate, too. This allows us not to have to solve this problem again when someone has a weird need to change `ssl_options`, for example. They can just supply the values in the URL connection string, and the values will be carried around for free.

src/radical/entk/appman/config.json
```diff
@@ -11,6 +9,7 @@
                          "db_cleanup"      : false },
     "pending_qs"      : 1,
     "completed_qs"    : 1,
-    "rmq_cleanup"     : true
+    "rmq_cleanup"     : true,
+    "rmq_url"         : "amqp://guest:guest@localhost:5672/%2F"
```
andre-merzky: Is the `%2F` intentional?

andre-merzky:
One thing which is a bit against our taste is that the appmgr constructor now has two routes of specifying, say, the RMQ hostname: via the explicit arg and via the URL.  I understand the rationale for the URL though, wrt. later possible additions like ssl params - but the API would be a bit more concise if we add just a `username` and a `password` argument, and construct the URL internally - what do you think?

src/radical/entk/appman/config.json
wilkinson: Yes, it's a URL-encoded "/", which is the default virtual host. I don't understand exactly why it needs to be URL-encoded, but that's what the Pika documentation said to do.

wilkinson:
I'm not sure why you say that the API would be made more concise by adding more arguments. Manual handling of each individual piece is not a prettier API; it's just a longer and more limited one. The only reason I know to use that strategy would be to support existing users' workflows.  Do you have any users who aren't in direct collaboration with you anyway?

My own preference was shown in the first pull request I sent, which was to replace the `hostname` and `port` stuff entirely because using a URL connection string is significantly more powerful. That strategy wasn't sensitive to existing users of the current API, however, so my second attempt extended the existing API. I agree that it's gross to have two routes of specifying things, and I don't like it at all. I'm an opinionated person, so I won't hesitate to tell you that the URL way is a lot better. Eventually there will be a user who wants to use a different virtual host or an SSL certificate file or something, and then you'll have to add more arguments again.

For example, I'm currently having problems with RADICAL-Pilot and MongoDB because of troubles with URL query parameters. That is exactly the problem that this strategy is trying to avoid. The URL-based strategy allows full access to the underlying Pika/RabbitMQ APIs without adding dozens of additional arguments to the `AppManager` constructor and everything downstream from it.

But honestly, I think it's inconsistent that RabbitMQ needs to be specified at all, because the MongoDB connections are set up implicitly by an environment variable. I'm not opposed to an implicit approach. I feel that they should match, which would also mean that there would only be one environment variable for RabbitMQ -- not two -- and it should be a URL, just like for `RADICAL_PILOT_DBURL`.

mturilli:
@wilkinson I agree with several of the points you made but EnTK is used in production by different groups on different resources. Any change in the API is potentially disruptive for ongoing research, experiments and publications. To avoid this, we extend the API and, in case, plan for a (long) period of deprecation before introducing a permanent change. 

Consistently, let's add a `credentials` parameter as previously indicated, using env variables for user name/passwd for a `pika.credentials.PlainCredentials` object to pass as credentials to `pika.ConnectionParameters`. Alongside this, let's keep your nice implementation of an alternative URL-based API. We will document both alternatives and see how users respond. Based on that response, we will decide whether to deprecate the existing API. 

Would that work for you? Happy to work together on this PR if useful.

wilkinson:
First, yes, I totally agree with approaches that are sensitive to existing users and their codes. Eventually, my goal is to build on top of EnTK myself eventually, and I won't want my code to come crashing down, either.

I only intended to argue that if I were designing an `AppManager` API with only one method of entry, I would choose a URL-based approach. Right now, I don't need to specify additional options with query parameters anyway, and basic credentials should be fine. I forget sometimes that the HPC world uses different conventions than I am used to. If you tell me that specifying four separate environment variables is the way to go, then I will support that. I assume the names would be something like `RMQ_HOSTNAME`, `RMQ_PORT`, `RMQ_USER`, and `RMQ_PASSWORD`, and a program might look like:
```python
import os
import pika
from radical.entk import AppManager

appman = AppManager(
    hostname=os.environ["RMQ_HOSTNAME"],
    port=os.environ["RMQ_PORT"],
    credentials=pika.credentials.PlainCredentials(
        user=os.environ["RMQ_USER"],
        password=os.environ["RMQ_PASSWORD"]))

# ...
```

Have I understood the desired interface correctly?

One thing we need to be sure to document is that special characters in the username and password need to become percent-encoded octets before being included in a URL string. We can work out the documentation later.

Also, how do you feel about the changes to the interfaces downstream of `AppManager`, where a `pika.connection.URLParameters` object is used to pass the credentials? I think it is equivalent to a `pika.connection.ConnectionParameters` object in functionality, but I understand if that needs to change. It is important to decide on that before updating the unit tests.

wilkinson:
I know we have a meeting in a few hours, but just to be clear, I'm not sure if the desired interface is the one shown above or perhaps something more like this:
```python
import os
from radical.entk import AppManager

appman = AppManager(
    hostname=os.environ["RMQ_HOSTNAME"],
    port=os.environ["RMQ_PORT"],
    username=os.environ["RMQ_USERNAME"],
    password=os.environ["RMQ_PASSWORD"])

# ...
```

andre-merzky:
Probably more like the second, so that we don't need to expose the user to `pika`, which should only be an internal implementation detail.  Thanks!

mturilli:
Hey @wilkinson, yes, I assumed this PR would be one of the main topic of today agenda. I agree with Andre, I would go with the second as it would just extend the current API.

wilkinson:
@andre-merzky, @mturilli:

After the discussion we had in our last meeting, I have now edited this pull request so that its only net changes are the following:
* `AppManager` accepts `username` and `password` parameters for RabbitMQ
* `AppManager` passes RabbitMQ connection parameters to downstream functions as a `pika.connection.ConnectionParameters` object instead of passing only the `hostname` and `port` parameters

This means that the following program will now work:
```python
import os
from radical.entk import AppManager

appman = AppManager(
    hostname=os.environ["RMQ_HOSTNAME"],
    port=os.environ["RMQ_PORT"],
    username=os.environ["RMQ_USERNAME"],
    password=os.environ["RMQ_PASSWORD"])

# ...
```

All of the URL stuff has been preserved in my [devel-url](https://github.com/wilkinson/radical.entk/tree/devel-url) branch, but it has been removed from this pull request and will be submitted for discussion as a different pull request in the future.

Obviously, the tests have not been updated yet, but I don't want to mess with that until I know for sure that this pull request is fine with everyone.

andre-merzky:
This looks good to me - thanks!  @mturilli ?

lee212:
@andre-merzky , LGTM as well. My thought was adding a test case for this e.g. if it is okay with special characters $, : or @. and if config.json does not contain 'username', 'password'.

andre-merzky:
@lee212 : Yeah, you are right on the missing tests of course.  I wonder though if those tests are doable w/o an RMQ instance configured for that test.  @wilkinson : would you mind looking into how easy it would be to add some tests cases, and what you would need from the infrastructure side to run those tests?

andre-merzky:
@wilkinson : ping - any input on tests?

wilkinson:
@andre-merzky Resolving #384 could be construed as a test, although not in the unit test sense. Right now, I have not been able to run the "Hello world" demo correctly, despite receiving excellent instructions from @mturilli. Because that issue uses the code from this pull request, I believe that we need to make sure the code passes that "test" before we worry about unit tests.

mturilli:
@wilkinson apologies for the delay on #384, these are two very busy weeks. I think we can at least discuss what unit test we may want to write for your PR, independent on whether the full EnTK+RP work on Summit?

andre-merzky:
ping @wilkinson ?

wilkinson:
@andre-merzky I would think that the way to unit-test this pull request would be to create an `AppManager` object by providing explicit, non-default connection parameters for RabbitMQ.

The `__init__` method calls the private method `_setup_mqs`, which means that simply creating an `AppManager` object will interact with RabbitMQ, which is all we need for a unit test. Unfortunately for our testing purposes, Pika will provide default parameters if any are missing. This means that specifying the defaults explicitly is not sufficient to test that connection parameters are handled correctly, and therefore it is necessary to provide explicit, non-default connection parameters. Thus, it appears that an instance of RabbitMQ with basic authentication enabled will be necessary for a unit test.

[CloudAMQP](https://www.cloudamqp.com/) offers a free plan that we might be able to use to test against, in the same way that you have used MongoLab in the past.

andre-merzky:
Merging now in order to not have it linger - thanks again for your patience!  I'll do the unit tests in a separate PR (can't push to your branch...).

PR #381, opened 2019-10-16
fix: #389

mturilli:
It installs clean, tests fail with an import issue.


andre-merzky:
Thanks, this looks good now.  The termination issues are still blocking before release though.

andre-merzky:
The Pythin3 conversion was merged already - this branch now contains only the fix for #389 for some reason, so I renamed the PR.  That diff looks good to me, my review approval holds for the merge.

Issue #384, opened 2019-10-29
ornl.summit does not seem to work from Summit login node


mturilli:
I am opening this ticket for @wilkinson, based on a mail thread now closed.

Stack: Unknown. @wilkinson please run `radical-stack` and add the output to this ticket.

@wilkinson is trying to run the following script from Summit login node:
```
import os
from radical.entk import AppManager, Pipeline, Stage, Task

# Create objects.

appman = AppManager(
    hostname=os.environ["RMQ_HOSTNAME"],
    port=os.environ["RMQ_PORT"],
    username=os.environ["RMQ_USERNAME"],
    password=os.environ["RMQ_PASSWORD"])

p = Pipeline()

s = Stage()

t = Task()

# Use the objects to model the workflow.

appman.resource_desc = {
    'resource': 'ornl.summit',
    'walltime': 10,
    'cpus': 1
}

t.name = 'my-first-task'
t.executable = '/bin/echo'
t.arguments = ['Hello world!']

s.add_tasks(t)

p.add_stages(s)

appman.workflow = set([p])

# Execute the workflow.

appman.run()
```

@wilkinson please add to this ticket a C&P of the trace the script returns, if any.

In attachment the logs/session of the run.
[ornl-summit-demo.tar.gz](https://github.com/radical-cybertools/radical.entk/files/3785723/ornl-summit-demo.tar.gz)

wilkinson:
Output from `radical-stack`:
```
  python               : 2.7.15
  pythonpath           : /autofs/nccs-svm1_sw/summit/.swci/0-core/opt/spack/20180914/linux-rhel7-ppc64le/gcc-4.8.5/py-virtualenv-16.0.0-ohsvxc5mf4aornhyrfp4ecea5bzcowon/lib/python2.7/site-packages:/sw/summit/xalt/1.1.4/site:/sw/summit/xalt/1.1.4/libexec
  virtualenv           : /ccs/home/seanwilk/entk-with-passwords

  radical.entk         : 0.72.0-v0.70.0-60-gc73e031@devel
  radical.pilot        : 0.73.1
  radical.saga         : 0.72.1
  radical.utils        : 0.72.0
```

mturilli:
Replication and debugging of the issue.

Setup:
```
module load python/2.7.15
module load py-virtualenv/16.0.0-py2
module load py-pip/10.0.1-py2
module load vim
module load py-setuptools/40.4.3-py2
[...]

PR #385, opened 2019-11-01
type conversion forced for rmq port

lee212:
Quick Fix for Preventing incorrect data type i.e. string to RabbitMQ port number, part of Issue #384


andre-merzky:
Thanks for the PR - we should have fixed this much earlier....

PR #394, opened 2019-12-11
Fix/issue 393

iparask:
It fixes issue #393.

PR #400, opened 2019-12-17
implement fix discussed in #399

### LINKED ISSUE #399: Workflow Processor does not advance stages
EnTK faces issues where the execution does not proceed when a stage is completed. I executed the getting started user guide example with the following stack:
```
  python               : 3.7.5
  pythonpath           :
  virtualenv           : rct_test

  radical.entk         : 0.90.6-v0.72.0-81-gad4d968c@fix-issue_393
  radical.pilot        : 0.90.8-v0.90.1-169-g1c35e226@devel
  radical.saga         : 0.90.6-bv0.72.1-67-g5deb253b@devel
  radical.utils        : 0.90.7-v0.90.3-176-g3cc0cd0@devel
```

When a stage is completed EnTK's workflow processor, check is there is a stage `post_exec` directive, [here](https://github.com/radical-cybertools/radical.entk/blob/fix/issue_393/src/radical/entk/appman/wfprocessor.py#L353).
If there is, it gets executed. If not, it checks whether the respective pipeline is completed.

This last statement is where the bug lies. When a stage post exec is executed the pipeline [goes to the new stage](https://github.com/radical-cybertools/radical.entk/blob/fix/issue_393/src/radical/entk/appman/wfprocessor.py#L412)
The pipeline does not move to the next stage if there is no post_exec, [no else](https://github.com/radical-cybertools/radical.entk/blob/fix/issue_393/src/radical/entk/appman/wfprocessor.py#L353) statement.

As a result, in the case of the getting started example, the stage number never increases, and the pipeline never get completed, and does not move in the `DONE` state. This in turn, does not allow the execution to finish.

Furthermore, I think that this bug would cause issues to any pipeline that has a stage without a post exec statement. The stage number would not be increased, as a result a completed stage would always be checked and the execution would stay there.

I suggest the following fix, [here](https://github.com/radical-cybertools/radical.entk/blob/fix/issue_393/src/radical/entk/appman/wfprocessor.py#L353):
```python
                    # Check if the current stage has a post-exec
                    # that needs to be executed
                    if stage.post_exec:
                        self._execute_post_exec(pipe, stage)
                    else:
                         pipe._increment_stage()
                    # If pipeline has completed, make state
                    # change
                    if pipe.completed:
```

I also suggest the fix to be implemented in the `fix/issue_393` branch which provides a fix for the termination calls.

andre-merzky:
This is based on branch `fix/issue_393` (I am still unsure about the branch state), and fixes #399 - but does *not* seem to fix termination issues.

### DIFF
--- a/src/radical/entk/appman/wfprocessor.py
+++ b/src/radical/entk/appman/wfprocessor.py
@@ -353,11 +353,18 @@ class WFprocessor(object):
                     if stage.post_exec:
                         self._execute_post_exec(pipe, stage)
 
-                    # If pipeline has completed, make state
-                    # change
-                    if pipe.completed:
+                    # if pipeline got suspended, advance state accordingly
+                    if pipe.state == states.SUSPENDED:
+                      self._advance(pipe, 'Pipeline', states.SUSPENDED)
 
+                    else:
+                        # otherwise perform normal stage progression
+                        pipe._increment_stage()
+
+                    # If pipeline has completed, advance state to DONE
+                    if pipe.completed:
                         self._advance(pipe, 'Pipeline', states.DONE)
+
 
                 # Found the pipeline and processed it -- no more
                 # iterations needed for the current task
@@ -404,15 +411,6 @@ class WFprocessor(object):
                         else:
                             self._advance(r_pipe, 'Pipeline', r_pipe.state)
 
-
-        if pipe.state == states.SUSPENDED:
-            self._advance(pipe, 'Pipeline', states.SUSPENDED)
-
-        else:
-            pipe._increment_stage()
-
-            if pipe.completed:
-                self._advance(pipe, 'Pipeline', states.DONE)
 
 
     # --------------------------------------------------------------------------


andre-merzky:
> I am very confused. If it is a branch out from `fix/issue_393` why is it pointed to `devel` and why this is the only change?
> 
> Can you make sure that you branched out from the correct branch and point the PR to the correct branch?

Fair enough - but I think the agreement was to merge the py3 changed toward devel to clean up, and then this PR would become small.  But either way, I'll reroute the PR!

andre-merzky:
My confusion is increasing: the diff between `fix/issue_393` and devel is zero, so how does that fix termination?

lee212:
`devel` branch includes `fix/issue_393` which explains the diff is zero. The base needs to be `devel`.

andre-merzky:
> `devel` branch includes `fix/issue_393` which explains the diff is zero. The base needs to be `devel`.

Shouldn't #394 be closed by then and  `fix/issue_393` be removed?  Either way, I'll revert the reverted base to devel :-)

iparask:
Because of this: https://github.com/radical-cybertools/radical.entk/commit/5357e08a84a257d1059b7dad1797d4741e153278

src/radical/entk/appman/wfprocessor.py
```diff
@@ -353,12 +353,19 @@ def _update_dequeued_task(self, deq_task):
...
-                    # If pipeline has completed, make state
-                    # change
-                    if pipe.completed:
+                    # if pipeline got suspended, advance state accordingly
+                    if pipe.state == states.SUSPENDED:
+                      self._advance(pipe, 'Pipeline', states.SUSPENDED)
+
+                    else:
+                        # otherwise perform normal stage progression
+                        pipe._increment_stage()
```
iparask: This will create a double pipeline increment with line 406. I am not sure how to fix it.

PR #412, opened 2020-01-28
Travis fix with documentation update for python 3

lee212:
TOC updated:
- installation
- user guide
- examples/getting started

Sections eliminated are:
- mongodb installation
- mongodb env configuration with mlab
   -  no env set leads to a default value to be loaded
- rabbitmq installation
- rabbitmq configuration
    - two.radical is exposed to use

### DIFF
--- a/docs/install.rst
+++ b/docs/install.rst
@@ -7,7 +7,7 @@ Installation
 Installing Ensemble Toolkit
 ===========================
 
-Installing Ensemble Toolkit using Pypi, or Git
+Installing Ensemble Toolkit using virtualenv
 ----------------------------------------------
 
 To install the Ensemble Toolkit, we need to create a virtual environment. 
@@ -15,11 +15,27 @@ Open a terminal and run:
 
 .. code-block:: bash
 
-        virtualenv $HOME/myenv -p python3.7
-        source $HOME/myenv/bin/activate
+        virtualenv $HOME/ve-entk -p python3.7
+        source $HOME/ve-entk/bin/activate
 
-*`-p` params indicates which python version you use, python3.6+ is required*
+- ``-p`` params indicates which python version you use, python3.6+ is required
+- python2 installation is available with the ``0.72.1`` version. Read more at
+  the troubleshooting_
 
+Activate virtualenv by:
+
+.. code-block:: bash
+
+        source $HOME/ve-entk/bin/activate
+
+.. note:: Activated env name is indicated in the prompt like: ``(ve-entk) username@two:~$``
+
+Deactivate the virtualenv, if you want to disengage. Your python won't
+recognize EnTK after deactivation.  To deactivate, run:
+
+.. code-block:: bash
+
+        deactivate
 
 It is suggested to use the released version of EnTK which you can install
 by executing the following command in your virtualenv:
@@ -29,8 +45,9 @@ by executing the following command in your virtualenv:
         pip install radical.entk
 
 
-To install a specific branch of EnTK, e.g. `devel`, you will need to clone the
-repository and checkout the branch. You can do so using the following commands:
+To install a specific branch of EnTK, e.g. `devel` instead of using pip
+installation, you will need to clone the repository and checkout the branch.
+You can do so using the following commands:
 
 .. code-block:: bash
 
@@ -49,9 +66,9 @@ printed.
 
         radical-stack
 
-          python               : 3.6.9
+          python               : 3.7.4
           pythonpath           :
-          virtualenv           : /home/hrlee/venv3/entk.py36
+          virtualenv           : /home/username/ve-entk
 
           radical.entk         : 1.0.0
           radical.pilot        : 1.0.0
@@ -67,8 +84,8 @@ Open a terminal and run (assuming you have PATH to point to ``conda``):
 
 .. code-block:: bash
 
-        conda create -n ve-entk python=3.7 -y
-        conda activate ve-entk
+        conda create -n conda-entk python=3.7 -y
+        conda activate conda-entk
 
 
 It is suggested to use the released version of EnTK which you can install
@@ -321,6 +338,8 @@ you want to access, you will have to get the certificates from the corresponding
 information is available in their user guide.
 
 
+.. _troubleshooting:
+
 Troubleshooting
 =======================
 
@@ -335,8 +354,35 @@ If virtualenv **is not** installed on your system, you can try the following.
         wget --no-check-certificate http://pypi.python.org/packages/source/v/virtualenv/virtualenv-16.7.9.tar.gz
         tar xzf virtualenv-16.7.9.tar.gz
 
-        python virtualenv-16.7.9/virtualenv.py $HOME/myenv -p python3.7
-        source $HOME/myenv/bin/activate
+        python virtualenv-16.7.9/virtualenv.py $HOME/ve-entk -p python3.7
+        source $HOME/ve-entk/bin/activate
+
+**Python 2 legacy installation**
+
+As of January 1, 2020, Python 2 support is terminated by the Python Software
+Foundation but the previous release of EnTK i.e. ``0.72.1`` allows to use Python 2.
+PyPI installation with virtualenv is:
+
+.. code-block:: bash
+
+        virtualenv $HOME/ve-entk-py2 -p python2.7
+        source $HOME/ve-entk-py2/bin/activate
+        pip install radical.entk==0.72.1
+
+```radical-stack``` confirms the versions of the radical cybertools:
+
+.. code-block:: bash
+
+        $ radical-stack
+
+          python               : 2.7.17
+          pythonpath           :
+          virtualenv           : /home/username/ve-entk-py2
+
+          radical.entk         : 0.72.1
+          radical.pilot        : 0.73.1
+          radical.saga         : 0.72.1
+          radical.utils        : 0.72.0
 
 .. comments
         **TypeError: 'NoneType' object is not callable**
--- a/docs/install.rst
+++ b/docs/install.rst
@@ -16,10 +16,9 @@ Open a terminal and run:
 .. code-block:: bash
 
         virtualenv $HOME/ve-entk -p python3.7
-        source $HOME/ve-entk/bin/activate
 
 - ``-p`` params indicates which python version you use, python3.6+ is required
-- python2 installation is available with the ``0.72.1`` version. Read more at
+- A legacy python2 installation is available with the ``0.72.1`` version. Hot fixes will be provided until Jul 2020. Read more at
   the troubleshooting_
 
 Activate virtualenv by:
@@ -45,7 +44,7 @@ by executing the following command in your virtualenv:
         pip install radical.entk
 
 
-To install a specific branch of EnTK, e.g. `devel` instead of using pip
+To install a specific branch of EnTK, e.g., `devel` instead of using pip
 installation, you will need to clone the repository and checkout the branch.
 You can do so using the following commands:
 
@@ -59,8 +58,8 @@ You can do so using the following commands:
 
 
 You can check the version of Ensemble Toolkit with the
-```radical-stack``` command-line tool. The current version should be
-printed.
+```radical-stack``` command-line tool. The currently installed version should
+be printed.
 
 .. code-block:: bash
 
@@ -97,8 +96,8 @@ by executing the following command in your conda env:
 
 
 You can check the version of Ensemble Toolkit with the
-```radical-stack``` command-line tool. The current version should be
-printed.
+```radical-stack``` command-line tool. The currently installed version should
+be printed.
 
 .. code-block:: bash
 
@@ -345,13 +344,13 @@ Troubleshooting
 
 **Missing virtualenv**
 
-This should return the version of the RADICAL-Pilot installation, e.g., `1.0.0`.
+This should return the version of the RCT installation, e.g., `1.0.0`.
 
 If virtualenv **is not** installed on your system, you can try the following.
 
 .. code-block:: bash
 
-        wget --no-check-certificate http://pypi.python.org/packages/source/v/virtualenv/virtualenv-16.7.9.tar.gz
+        wget --no-check-certificate https://pypi.python.org/packages/source/v/virtualenv/virtualenv-16.7.9.tar.gz
         tar xzf virtualenv-16.7.9.tar.gz
 
         python virtualenv-16.7.9/virtualenv.py $HOME/ve-entk -p python3.7
@@ -360,7 +359,7 @@ If virtualenv **is not** installed on your system, you can try the following.
 **Python 2 legacy installation**
 
 As of January 1, 2020, Python 2 support is terminated by the Python Software
-Foundation but the previous release of EnTK i.e. ``0.72.1`` allows to use Python 2.
+Foundation but the previous release of EnTK i.e. ``0.72.1`` allows to use Python 2.7.
 PyPI installation with virtualenv is:
 
 .. code-block:: bash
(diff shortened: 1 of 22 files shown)


docs/install.rst
```diff
@@ -70,122 +97,150 @@ by executing the following command in your conda env:
...
+                RUN apt-get update \
+                && apt-get install wget curl python python-dev python-pip python-virtualenv bzip2 -y \
+                && virtualenv ~/ve-entk \
+                && . ~/ve-entk/bin/activate \
+                && pip install radical.entk
 
-Once you execute the container, the default path will be /root (of the container).
-The EnTK virtualenv exists at ~/ve-entk (inside the container). This is useful
-to know as the examples exist inside the virtualenv.
+        You can also download the Dockerfile :download:`here <./misc/Dockerfile>`.
```
mturilli: I would update this dockerfile as suggested above.

docs/install.rst
```diff
@@ -70,122 +97,150 @@ by executing the following command in your conda env:
...
+                docker run -t -i entk
 
+        Once you execute the container, the default path will be /root (of the container).
+        The EnTK virtualenv exists at ~/ve-entk (inside the container). This is useful
+        to know as the examples exist inside the virtualenv.
 
-Installing rabbitmq
-===================
+        You can check the version of Ensemble Toolkit with the
+        ```radical-entk-version``` command-line tool. The current version should be
```
mturilli: current -> currently installed?

docs/install.rst
```diff
@@ -70,122 +97,150 @@ by executing the following command in your conda env:
...
-Installing rabbitmq as a system process (sudo privileges required)
-------------------------------------------------------------------
+        .. code-block:: bash
 
-Ensemble Toolkit relies on RabbitMQ for message transfers. Installation
-instructions can be found at <https://www.rabbitmq.com/download.html>. At
-the end of the installation run ```rabbitmq-server``` to start the server.
-RabbitMQ needs to be installed on the same machine as EnTK is installed.
+                radical-entk-version
+                0.70.0
```
mturilli: I would update to 1.0 after updating the docker file to p3.

docs/install.rst
```diff
@@ -70,122 +97,150 @@ by executing the following command in your conda env:
...
+Installation instructions can be found at
+<https://www.rabbitmq.com/download.html>. At the end of the installation run
+```rabbitmq-server``` to start the server.
+
+The following configuration defines a default server and port number to communicate.
 
 .. code-block:: bash
 
-        rabbitmq-server -detached
+        export RMQ_HOSTNAME=two.radical-project.org; export RMQ_PORT=33239
```
mturilli: I would anonymize this, especially because at the moment we do not password-protect it. Something like:
```
         export RMQ_HOSTNAME=<hostname>; export RMQ_PORT=port
```
I would also add documentation about password protected hosts. See Sean patch and relative documentation.

docs/install.rst
```diff
@@ -70,122 +97,150 @@ by executing the following command in your conda env:
...
-docker instance of rabbitmq. Assuming you have docker installed, you can
-download and run the rabbitmq instance using the following command:
+        Installing rabbitmq as a system process (sudo privileges required)
+        ------------------------------------------------------------------
 
-.. code-block:: bash
+        Ensemble Toolkit relies on RabbitMQ for message transfers. Installation
+        instructions can be found at <https://www.rabbitmq.com/download.html>. At
+        the end of the installation run ```rabbitmq-server``` to start the server.
+        RabbitMQ needs to be installed on the same machine as EnTK is installed.
```
mturilli: I would eliminate `RabbitMQ needs to be installed on the same machine as EnTK is installed.` as it is not generally true.

docs/install.rst
```diff
@@ -70,122 +97,150 @@ by executing the following command in your conda env:
...
-.. note::
-   If you are using Docker to install both EnTK and RabbitMQ, they should run
-   as two different containers. You can set the RMQ_PORT in the EnTK container
-   accordingly.
 
-Installation Video
-==================
+        The '-P' argument auto maps new ports from localhost to the ports expected by
+        rabbitmq. This is useful if you want to have multiple EnTK scripts running as
+        you would require multiple rabbitmq instances.
```
mturilli: I would eliminate `This is useful if you want to have multiple EnTK scripts running as you would require multiple rabbitmq instances` as I think this is not true anymore. Multiple EnTK should be able to share the same rabbitmq server.

docs/install.rst
```diff
@@ -70,122 +97,150 @@ by executing the following command in your conda env:
...
+                vivek@two:~$ docker run -d --name rabbit-1 -P rabbitmq:3
+                fb8ee8bfd822656a6338b7c19fa6a9641944f8bf5de5c1414fb78d049fdffc42
+                vivek@two:~$ docker ps
+                CONTAINER ID        IMAGE               COMMAND                  CREATED             STATUS              PORTS                                                                                                 NAMES
+                fb8ee8bfd822        rabbitmq:3          "docker-entrypoint..."   9 seconds ago       Up 7 seconds        0.0.0.0:32777->4369/tcp, 0.0.0.0:32776->5671/tcp, 0.0.0.0:32775->5672/tcp, 0.0.0.0:32774->25672/tcp   rabbit-1
+
+
+        Interactions between RabbitMQ and EnTK are done through port 5672 by default.
+        For the above docker instance, we need to use port 32775. In your EnTK scripts,
+        while creating the AppManager, you need to specify port=32775.
```
mturilli: I would add: `or read the port number from RMQ_PORT`.

docs/install.rst
```diff
@@ -70,122 +97,150 @@ by executing the following command in your conda env:
...
+           If you are using Docker to install both EnTK and RabbitMQ, they should run
+           as two different containers. You can set the RMQ_PORT in the EnTK container
+           accordingly.
+
+        Installation Video
+        ==================
+
+        .. raw:: html
+
+                <video controls width="800" src="_static/entk_installation_get_started.mp4"></video>
```
mturilli: Is this current? Should we update it?

docs/install.rst
```diff
@@ -197,54 +252,56 @@ GSISSH, but it requires (a) a MongoDB server and (b) a properly set-up
...
 
-The MongoDB server is used to store and retrieve operational data during the
-execution of an application using RADICAL-Pilot. The MongoDB server must
-be reachable on **port 27017** from **both**, the host that runs the
-Ensemble Toolkit application and the host that executes the MD tasks, i.e.,
-the HPC cluster (see blue arrows in the figure above). In our experience,
-a small VM instance (e.g., Amazon AWS) works exceptionally well for this.
+        The MongoDB server is used to store and retrieve operational data during the
+        execution of an application using RADICAL-Pilot. The MongoDB server must
+        be reachable on **port 27017** from **both**, the host that runs the
```
mturilli: **port 27017** -> on a specified port? I do not think we need to use 27017 right?

docs/install.rst
```diff
@@ -197,54 +252,56 @@ GSISSH, but it requires (a) a MongoDB server and (b) a properly set-up
...
+        Ensemble Toolkit application and the host that executes the MD tasks, i.e.,
+        the HPC cluster (see blue arrows in the figure above). In our experience,
+        a small VM instance (e.g., Amazon AWS) works exceptionally well for this.
```
mturilli: I would delete `exceptionally`.

docs/install.rst
```diff
@@ -197,54 +252,56 @@ GSISSH, but it requires (a) a MongoDB server and (b) a properly set-up
...
 
-**Install your own MongoDB**
+        **Install your own MongoDB**
 
-Once you have identified a host that can serve as the new home for MongoDB,
-installation is straight forward. You can either install the MongoDB
-server package that is provided by most Linux distributions, or
-follow the installation instructions on the MongoDB website:
+        Once you have identified a host that can serve as the new home for MongoDB,
+        installation is straight forward. You can either install the MongoDB
```
mturilli: straight forward -> straightforward

docs/install.rst
```diff
@@ -197,54 +252,56 @@ GSISSH, but it requires (a) a MongoDB server and (b) a properly set-up
...
+        server package that is provided by most Linux distributions, or
+        follow the installation instructions on the MongoDB website:
 
-http://docs.mongodb.org/manual/installation/
+        http://docs.mongodb.org/manual/installation/
 
-**MongoDB-as-a-Service**
+        **MongoDB-as-a-Service**
```
mturilli: Eliminate the whole `**MongoDB-as-a-Service**` 'section' as we do not support mlab anymore.

docs/install.rst
```diff
@@ -268,8 +325,8 @@ Now you can login to the machine by ``ssh machine1``.
 Source: http://nerderati.com/2011/03/17/simplify-your-life-with-an-ssh-config-file/
 
 
-Setup GSISSH Access to a machine
----------------------------------
+Setup GSISSH Access to HPC resources
+------------------------------------
 
 Setting up GSISSH access to a machine is a bit more complicated. We have documented the steps to setup GSISSH on
```
mturilli: I would start the paragraph with: `Currently, GSISSH is unsupported but several HPC platforms still require it.`

docs/install.rst
```diff
@@ -281,33 +338,63 @@ you want to access, you will have to get the certificates from the corresponding
 information is available in their user guide.
```
mturilli: Are these guides to install SSH still working? I suspect both are way outdated.

docs/install.rst
```diff
@@ -281,33 +338,63 @@ you want to access, you will have to get the certificates from the corresponding
...
 
+.. _troubleshooting:
+
 Troubleshooting
 =======================
 
 **Missing virtualenv**
 
-This should return the version of the RADICAL-Pilot installation, e.g., `0.X.Y`.
+This should return the version of the RADICAL-Pilot installation, e.g., `1.0.0`.
```
mturilli: Why RADICAL-Pilot?

docs/install.rst
```diff
@@ -281,33 +338,63 @@ you want to access, you will have to get the certificates from the corresponding
...
 
 If virtualenv **is not** installed on your system, you can try the following.
 
 .. code-block:: bash
 
-        wget --no-check-certificate https://pypi.python.org/packages/source/v/virtualenv/virtualenv-1.9.tar.gz
-        tar xzf virtualenv-1.9.tar.gz
+        wget --no-check-certificate http://pypi.python.org/packages/source/v/virtualenv/virtualenv-16.7.9.tar.gz
```
mturilli: I think this command is outdated. See `https://virtualenv.pypa.io/en/latest/installation/`. I get: 
```
$ wget --no-check-certificate http://pypi.python.org/packages/source/v/virtualenv/virtualenv-16.7.9.tar.gz
--2020-01-29 10:49:32--  http://pypi.python.org/packages/source/v/virtualenv/virtualenv-16.7.9.tar.gz
Resolving pypi.python.org (pypi.python.org)... 2a04:4e42:1b::223, 151.101.12.223
Connecting to pypi.python.org (pypi.python.org)|2a04:4e42:1b::223|:80... connected.
HTTP request sent, awaiting response... 403 SSL is required
2020-01-29 10:49:32 ERROR 403: SSL is required.
```

docs/install.rst
```diff
@@ -281,33 +338,63 @@ you want to access, you will have to get the certificates from the corresponding
...
+        tar xzf virtualenv-16.7.9.tar.gz
+
+        python virtualenv-16.7.9/virtualenv.py $HOME/ve-entk -p python3.7
+        source $HOME/ve-entk/bin/activate
+
+**Python 2 legacy installation**
+
+As of January 1, 2020, Python 2 support is terminated by the Python Software
+Foundation but the previous release of EnTK i.e. ``0.72.1`` allows to use Python 2.
```
mturilli: allows to use Python 2 -> allows to use Python 2.7

docs/install.rst
```diff
@@ -281,33 +338,63 @@ you want to access, you will have to get the certificates from the corresponding
...
 
-        python virtualenv-1.9/virtualenv.py $HOME/myenv
-        source $HOME/myenv/bin/activate
+          radical.entk         : 0.72.1
+          radical.pilot        : 0.73.1
+          radical.saga         : 0.72.1
+          radical.utils        : 0.72.0
 
-**TypeError: 'NoneType' object is not callable**
+.. comments
```
mturilli: I would eliminate this comment as we do not support python 2 anymore and above, we have mentioned only python 2.7 as the legacy version supported.

docs/user_guide/get_started.rst
```diff
@@ -16,14 +16,16 @@ You can download the complete code discussed in this section :download:`here <..
 your virtualenv under ``share/radical.entk/user_guide/scripts``.
 
 
-Use pre-configured MongoDB
-==========================
+.. comments
 
-For the purposes of this user guide, we have a MongoDB setup to use. Please run the following command to use it:
+        Use pre-configured MongoDB
```
mturilli: I would eliminate the whole 'section' `Use pre-configured MongoDB` as we do not support mlab anymore.

mturilli:
Note that Travis is broken. All the tests fail and it is still using Python 2.7. This need to be fixed before our next release on Feb 4th.

docs/install.rst
```diff
@@ -7,17 +7,34 @@ Installation
...
 ----------------------------------------------
 
 To install the Ensemble Toolkit, we need to create a virtual environment. 
 Open a terminal and run:
 
 .. code-block:: bash
 
-        virtualenv $HOME/myenv
-        source $HOME/myenv/bin/activate
+        virtualenv $HOME/ve-entk -p python3.7
```
mturilli: While working on Frontera, I saw that the suggested way to create a virtualenv with >= python3.3 is: `python3 -m venv $HOME/ve-entk` (https://docs.python.org/3/library/venv.html) should we move to this?

docs/install.rst
```diff
@@ -27,26 +44,35 @@ by executing the following command in your virtualenv:
...
+To install a specific branch of EnTK, e.g., `devel` instead of using pip
+installation, you will need to clone the repository and checkout the branch.
+You can do so using the following commands:
 
 .. code-block:: bash
 
         git clone https://github.com/radical-cybertools/radical.entk.git
         cd radical.entk
-        git checkout <branch-name>
+        git checkout devel
```
mturilli: I would leave <branch-name> so to remark the idea that this is a general purpose way to checkout whatever branch the user needs. This (used to?) happen when a fix was made specifically for a user before release.

docs/install.rst
lee212: These two are different libraries, venv is a standard lib and virtualenv is an external lib from pypi although I don't see much difference in the way of building a separate environment (directory e.g. ve-entk for site-packages). I am not against replacing virtualenv to -m venv. Let me know what you think.

andre-merzky: Maybe, in the few tests I did it worked.  But note that `venv` does behave different than virtualenv, and misses a couple of features, so maybe we use it a while and test before we recommend / document it as the default way?  If it seems stable, we can also consider using it in the agent bootstrapper.

mturilli: Fair, they are indeed different so I agree with the suggestion to test it before for a while. On Frontera and other XSEDE resource(s) (do not remember now which one, might have been stampede2/comet), with virtualenv, I was not able to update pip while I was able with venv. That was a roadblock where pip was outdated enough to fail installing RCT. Eliminating virtualenv has also the benefit to reduce of one package our dependences (a package that is not well maintaned sometimes on the HPC machines, requiring user-level installation, something that happen to me).

andre-merzky: Ack on all points...

docs/install.rst
lee212: Make sense. I will revert it back.

lee212:
> Note that Travis is broken. All the tests fail and it is still using Python 2.7. This need to be fixed before our next release on Feb 4th.

Oh I see that I forgot to merge #406 by Ioannis which updates travis CI to use python 3.6. I added it now.

lee212:
Travis is broken, not only python version but also test failures. This needs to be separated with doc update. I can create new PR for travis but hope to close this PR first for publishing new readthedocs.

lee212:
Changed base to `devel` because this PR became Travis repair along with documentation update for python3.

lee212:
I confirm that most of failed cases before have been fixed now. RP related tests may fail but it will be resolved once #radical-cybertools/radical.pilot/pull/2041 is merged to `master` and released.

PR #435, opened 2020-03-07
early assignment of UIDs

andre-merzky:
This PR moves the UID creation for pipelines, tasks and stages into the respective constructors.  It also moves the assignment of `parent_stage` and `parent_pipeline` attributes into the `add_task` and `add_stage` methods which establishes that parental relationship.

This PR probably needs some thorough testing to ensure that no code depends on the late UID assignment.

### DIFF
--- a/src/radical/entk/pipeline/pipeline.py
+++ b/src/radical/entk/pipeline/pipeline.py
@@ -9,15 +9,16 @@ from collections import Iterable
 class Pipeline(object):
 
     """
-    A pipeline represents a collection of objects that have a linear temporal execution order.
-    In this case, a pipeline consists of multiple 'Stage' objects. Each ```Stage_i``` can execute only
-    after all stages up to ```Stage_(i-1)``` have completed execution.
+    A pipeline represents a collection of objects that have a linear temporal
+    execution order.  In this case, a pipeline consists of multiple 'Stage'
+    objects. Each ```Stage_i``` can execute only after all stages up to
+    ```Stage_(i-1)``` have completed execution.
 
     """
 
     def __init__(self):
 
-        self._uid = None
+        self._uid  = ru.generate_id('pipeline.%(item_counter)04d', ru.ID_CUSTOM)
         self._name = None
 
         self._stages = list()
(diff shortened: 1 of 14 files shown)


andre-merzky:
@lee212 : will you find the time to review this PR before the next devel call?  If not, please assign a different reviewer - thanks!

src/radical/entk/pipeline/pipeline.py
```diff
@@ -9,15 +9,16 @@
...
+    execution order.  In this case, a pipeline consists of multiple 'Stage'
+    objects. Each ```Stage_i``` can execute only after all stages up to
+    ```Stage_(i-1)``` have completed execution.
 
     """
 
     def __init__(self):
 
-        self._uid = None
+        self._uid  = ru.generate_id('pipeline.%(item_counter)04d', ru.ID_CUSTOM)
```
lee212: Isn't the namespace `(ns=sid)` needed?

lee212: Test runs of `get_started.py` shows incremented uid indexes even though new session created.

- First run:
```
Update: pipeline.0000 state: SCHEDULING
Update: pipeline.0000.stage.0000 state: SCHEDULING
Update: pipeline.0000.stage.0000.my-first-task state: SCHEDULING
...
```

- Second run:
```
Update: pipeline.0001 state: SCHEDULING
Update: pipeline.0001.stage.0001 state: SCHEDULING
Update: pipeline.0001.stage.0001.my-first-task state: SCHEDULING
```

andre-merzky: Thanks, good catch!  the namespace is only really needed if the same type of entities get created in different components of the stack - which is not the case in RE.  But I indeed had to change from `item_counter` (which persists for the item type) to `counter` (which restarts for every application instance).  I made that consistent for all uid calls now.

First run:
```
re.session.rivendell.merzky.018351.0016
pipeline.0000 stage.0000 task.0000
pipeline.0000 stage.0001 task.0001
pipeline.0000 stage.0001 task.0002
```

Second run:
```
re.session.rivendell.merzky.018351.0017
pipeline.0000 stage.0000 task.0000
pipeline.0000 stage.0001 task.0001
pipeline.0000 stage.0001 task.0002
```

PR #449, opened 2020-05-27
update private attribute, _shared_data

lee212:
`shared_data` wasn't updated to the private attribute `_shared_data` of the class `AppManager`, although it is updated in the `_rmgr` if exists. This hotfix applies the change of the attribute in the class as well to address the issue from: https://github.com/radical-collaboration/hpc-workflows/issues/119


lee212:
> This PR is a great candidate for a test.

@iparask what steps to take for this?

PR #450, opened 2020-05-29
Fix/docs for data staging and post_exec

lee212:
Descriptions are updated with examples to provide a better understanding.

### DIFF
--- a/docs/adv_examples/adapt_ta.rst
+++ b/docs/adv_examples/adapt_ta.rst
@@ -4,31 +4,61 @@
 Adaptive applications: Task-attribute
 *************************************
 
-We encourage you to take a look at the Pipeline of Ensembles example on the next page and compare it with the above
-pattern.
+We encourage you to take a look at the Pipeline of Ensembles example on the
+next page and compare it with the above pattern.
 
 .. note:: The reader is assumed to be familiar with the :ref:`PST Model <app_model>` and to have read through the :ref:`introduction` of Ensemble Toolkit.
 
 .. note:: This chapter assumes that you have successfully installed Ensemble Toolkit, if not see :ref:`Installation`.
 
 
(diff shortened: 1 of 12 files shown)


docs/adv_examples/adapt_ta.rst
```diff
@@ -4,31 +4,61 @@
...
-can only be performed once all tasks of a Stage are completed as the concurrent tasks cannot be interrupted by design.
-The post execution property of a Stage requires 3 function handles: a function that returns a boolean, a function that is
-executed when the boolean result is True, and a function that is executed when the boolean result is False.
+For any adaptive capability within a Pipeline, we need to use the post
+execution property of a Stage object. Decisions can only be performed once all
+tasks of a Stage are completed as the concurrent tasks cannot be interrupted by
+design.  The post execution property of a Stage requires a callable function
+that determines next stages of a workflow, and a (True) function is called to
+add a new stage or continue when a condition is satisfied, of not a (False)
+function is called to stop.
```
andre-merzky, line 26: I think this is not true anymore, and the documentation on the `stage.post_exec` property is outdated: the post_exec property now takes a single callable and calls it (see `_execute_post_exex` in `wf_processor.py` [here](https://github.com/radical-cybertools/radical.entk/blob/devel/src/radical/entk/appman/wfprocessor.py#L382).

lee212: @andre-merzky thanks for the comment. I removed the sentence and yes, the updated document was trying to explain it now takes a single callable object instead of a python dictionary but it shows that decision making is still doable by including other callable objects.

lee212: @andre-merzky, let me try to remove True/False style functions and make single callable again.

lee212: Thanks for the quick update.

PR #461, opened 2020-06-26
Add warning for KeyboardInterrupt exception

mturilli:
This add a warning after the first example about the incorrect KeyboardInterrupt exception.


mturilli:
@lee212 I see many tests fail with meaningful and therefore worrying failures. This should all be addressed before the next release planned for July 7th. As a general rule, we should not release with failing tests.

PR #466, opened 2020-06-30
Enhancement/no sync with master

lee212:
This reduces overhead in the `cud_create` event from `0.16` second to `0.0029` second per task, 56x speed up.

### DIFF
--- a/src/radical/entk/appman/appmanager.py
+++ b/src/radical/entk/appman/appmanager.py
@@ -807,13 +807,12 @@ class AppManager(object):
                         if completed_task.path:
                             task.path = str(completed_task.path)
 
-                        """mq_channel.basic_publish(
-                                exchange='',
-                                routing_key=reply_to,
-                                properties=pika.BasicProperties(
-                                    correlation_id=corr_id),
-                                body='%s-ack' % task.uid)
-                        """
+                        # mq_channel.basic_publish(
+                        #        exchange='',
+                        #        routing_key=reply_to,
+                        #        properties=pika.BasicProperties(
+                        #            correlation_id=corr_id),
+                        #        body='%s-ack' % task.uid)
 
                         state = msg['object']['state']
                         self._prof.prof('pub_ack_state_%s' % state,
--- a/src/radical/entk/appman/appmanager.py
+++ b/src/radical/entk/appman/appmanager.py
@@ -16,6 +16,7 @@ from .. import exceptions as ree
 
 from ..pipeline    import Pipeline
 from ..task        import Task
+from radical.entk  import states
 from ..utils       import write_session_description
 from ..utils       import write_workflows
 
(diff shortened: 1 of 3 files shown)


andre-merzky:
Well, yes, I can imagine it speeds things up - but it also removes the `ack` semantics from the RMQ message protocol.  I think that was introduced to increase reliability of the communiation channel.  I am not sure about the tradeoffs - @mturilli ?

mturilli:
I had to try to find notes about this but did not manage to find anything conclusive. I do remember we had an issue with eliminating the ack but I cannot say more. @lee212 , how did you test this?

lee212:
It's been a while but some explanation would help. The task manager and app manager continue to have `ack` semantics with this PR. They exchange messages to sync through two channels, `sync-to-tmgr` and `sync-to-cb`, these have not been touched.  The `ack` that I removed was located after all important parts, e.g. `_task_update` but sends an acknowledgment from the app manager to the task manager although it doesn't take any further action. I find this is redundant and doesn't break or reduce reliability. 

I also like to note that this PR is agreed to merge during our discussion. Would it be okay to merge?

andre-merzky:
Thanks for explaining!  It looks good to me to merge, if @mturilli has no further concerns.

src/radical/entk/appman/appmanager.py
```diff
@@ -792,19 +793,27 @@ def _task_update(self, msg, reply_to, corr_id, mq_channel, method_frame):
...
                             task.path = str(completed_task.path)
 
-                        mq_channel.basic_publish(
+                        """mq_channel.basic_publish(
                                 exchange='',
                                 routing_key=reply_to,
                                 properties=pika.BasicProperties(
                                     correlation_id=corr_id),
                                 body='%s-ack' % task.uid)
+                        """
```
andre-merzky, line 816: This should be a comment, not an string.

src/radical/entk/execman/base/task_manager.py
```diff
@@ -158,7 +158,7 @@ def _sync_with_master(self, obj, obj_type, channel, queue):
         reply_queue = '-'.join(list(reversed(qname)))
         reply_queue = sid + '-' + reply_queue
 
-        while True:
+        while False:
```
andre-merzky, line 161: This should also be commented out, if possible with the explanation you wrote in the PR comments.

andre-merzky:
I added two comments on the code, please address them before merge (should be quick).  Thanks!

src/radical/entk/appman/appmanager.py
lee212: thanks, andre. it is updated by 9e148bc

src/radical/entk/execman/base/task_manager.py
lee212: Thanks again, 9e148bc, 8ec40a5

PR #468, opened 2020-07-07
Fix/issue 465

lee212:
missing examples to update with `name` in data staging, instead of `uid`.


lee212:
mixed up. let me re-create new PR.

Issue #462, opened 2020-06-26
Several unit tests fail


mturilli:
Unit tests should all pass at release time. Currently, I see the following tests failing:
```
[gw0] [  6%] FAILED tests/test_component/test_amgr.py::test_amgr_initialization 
[gw1] [ 19%] FAILED tests/test_component/test_pipeline.py::test_pipeline_pass_uid 
[gw2] [ 21%] FAILED tests/test_component/test_amgr.py::test_state_order 
[gw1] [ 22%] FAILED tests/test_component/test_rmgr_2.py::test_rmgr_rp_get_resource_allocation_state 
[gw2] [ 26%] FAILED tests/test_component/test_pipeline.py::test_pipeline_to_dict 
[gw2] [ 38%] FAILED tests/test_component/test_stage.py::test_stage_to_dict 
[gw2] [ 46%] FAILED tests/test_component/test_stage.py::test_stage_pass_uid 
[gw2] [ 54%] FAILED tests/test_component/test_task.py::test_task_initialization 
[gw1] [ 55%] FAILED tests/test_component/test_rmgr_2.py::test_rmgr_rp_resource_request 
[gw1] [ 56%] FAILED tests/test_component/test_stage.py::test_stage_initialization 
[gw2] [ 60%] FAILED tests/test_component/test_task.py::test_task_to_dict eat 
[gw1] [ 80%] FAILED tests/test_component/test_tmgr_2.py::test_tmgr_rp_tmgr  
[gw2] [ 85%] FAILED tests/test_component/test_tproc_rp_2.py::test_create_task_from_cu 

[gw2] [  4%] FAILED tests/test_integration/test_rmgr.py::test_rmgr_base_initialization 
[gw2] [ 36%] FAILED tests/test_integration/test_rmgr.py::test_rmgr_rp_initialization 
[gw0] [ 44%] FAILED tests/test_integration/test_integration_local.py::test_integration_local 
[gw1] [ 48%] FAILED tests/test_integration/test_post_exec.py::test_stage_post_exec 
[gw2] [ 72%] FAILED tests/test_integration/test_rmgr_2.py::test_rmgr_rp_get_resource_allocation_state 
[gw1] [ 76%] FAILED tests/test_integration/test_tproc_rp.py::test_create_task_from_cu 
[gw0] [ 88%] FAILED tests/test_integration/test_shared_data.py::test_shared_data 
[...]

PR #471, opened 2020-07-29
Fix/failed tests

lee212:
Hopefully addresses failed tests #462


lee212:
Travis is passed. codecov is up nenxt.

andre-merzky:
I'm afraid the tests are a false positive: they time out after 8% completion :-(
https://travis-ci.com/github/radical-cybertools/radical.entk/jobs/370765627

lee212:
@andre-merzky I guess this is ready to merge.

andre-merzky:
From my end, yes.  Let's wait though until @mturilli finds time for the review.  Thanks @lee212

tests/test_component/test_amgr.py
```diff
@@ -22,6 +23,8 @@
 
 host     =     os.environ.get('RMQ_HOSTNAME', 'localhost')
 port     = int(os.environ.get('RMQ_PORT',      5672))
+username = os.environ.get('RMQ_USERNAME', 'guest')
+password = os.environ.get('RMQ_PASSWORD', 'guest')
 user     = 'guest'
```
mturilli, line 28: Can `user` and `passwd` guest be eliminated now that we use no defaults?

tests/test_component/test_amgr.py
```diff
@@ -464,6 +481,9 @@ def test_amgr_synchronizer():
     proc.start()
     proc.join()
 
+    # Wait for AppManager to finish the message exchange
+    time.sleep(5)
```
mturilli, line 485: Why do we need this? Is this only for the test or have we introduced the need of a wait to avoid a race condition in AppManager?

tests/test_component/test_tmgr.py
```diff
@@ -27,6 +27,8 @@
 
 hostname = os.environ.get('RMQ_HOSTNAME', 'localhost')
 port     = int(os.environ.get('RMQ_PORT', 5672))
+username = os.environ.get('RMQ_USERNAME', 'guest')
+password = os.environ.get('RMQ_PASSWORD', 'guest')
```
mturilli, line 32: Is this preset in an ephemeral RBMQ we setup on Travis? If not, do we still use 'guest'?

tests/test_component/test_tmgr_advance.py
```diff
@@ -18,8 +18,12 @@ def func(obj, obj_type, new_state, queue1):
 
     hostname = os.environ.get('RMQ_HOSTNAME', 'localhost')
     port = int(os.environ.get('RMQ_PORT', 5672))
+    username = os.environ.get('RMQ_USERNAME', 'guest')
+    password = os.environ.get('RMQ_PASSWORD', 'guest')
```
mturilli, line 23: Same comment as before. I will not repeat it for every test to avoid clutter.

tests/test_issues/test_diff_rmq.py
```diff
@@ -1,11 +1,11 @@
 from radical.entk import Pipeline, Stage, Task, AppManager
-from radical.entk.exceptions import *
 import pytest
-import os
 
 hostname = 'two.radical-project.org'
 port = 33142
```
mturilli, line 6: Two is dead, we should clean the code up.

tests/test_component/test_amgr.py
lee212: Yes, 
1c7362a removed the default values

tests/test_component/test_amgr.py
lee212: This test creates a thread/process to check *state* sync between AppManager and TaskManager. The problem is that it is too fast to terminate the thread in the test, not in the AppManager logic.

lee212: I removed this `time.sleep` at 401a458

PR #483, opened 2020-10-07
reuse rmq conn in unit_state_cb

lee212:
Single connection per thread is recommended and the `unit_state_cb()` is now reusing `mq_conn` instead of open/close.
This will reduce overhead in processing a large number of tasks as single connection/channel is used for multiple callbacks.


andre-merzky:
Wait, RE created a new connection *per task* -- no *per state update*??  WTH...  Glad you fixed that...

AymenFJA (approved):

PR #491, opened 2020-10-22
Doc string for a method

### LINKED ISSUE #401: Include a docstring and comment in _execute_post_exec in WFProcessor
The `_execute_post_exec` in WFProcessor executes a complicated feature, but there is no docstring or comments that explain what it does.

iparask:
Closes #401.

I added a small doc string. After reading the method I found that it is self explanatory what is happening. So there is no need for something significant


lee212 (approved):
LGTM

Issue #481, opened 2020-10-03
Align thread/processes definitions between EnTK and RP


mturilli:
The current differences create unnecessary confusion.

iparask:
Hello @mturilli, I just checked the attributes naming and it seems to be the same. From RP, we have:

```
CPU_PROCESSES          = 'cpu_processes'
CPU_PROCESS_TYPE       = 'cpu_process_type'
CPU_THREADS            = 'cpu_threads'
CPU_THREAD_TYPE        = 'cpu_thread_type'

GPU_PROCESSES          = 'gpu_processes'
GPU_PROCESS_TYPE       = 'gpu_process_type'
GPU_THREADS            = 'gpu_threads'
GPU_THREAD_TYPE        = 'gpu_thread_type'
```

EnTK introduces dictionaries for CPU and GPU requirements. I can either change the dictionary keys to much RPs or make a API change on EnTK to much that of RP. I think the first option is better and I will introduce a deprecation message for users.

iparask:
Closing as PR #492 is merged

PR #492, opened 2020-10-22
Aligning thread/processes definitions between EnTK and RP

iparask:
This PR aligns the definition between RP and EnTK for threads and processes definition. It still keeps the old keys and produces a deprecation warning once


andre-merzky:
Hmm, this is a difficult one: I am all for aligning the task descriptions - but then again, I am sure @mturilli  will not like that this changes the API (even with deprecation warning).  Also, it misses the target a bit: RP [uses](https://github.com/radical-cybertools/radical.pilot/blob/devel/src/radical/pilot/compute_unit_description.py#L20) `cpu_threads`, not `cpu_threads_per_process` for example.

Last but not least though, I am not happy with the RP task description either.  For example, we specify `gpu_processes`, but that is actually interpreted as `gpus_per_process`.  That is something we need to fix anyway, so an iteration of the task description is upcoming (also for other reasons, like heterogeneous ranks, prepared environments, containers, ...).

Summary: I am glad you look into this, but I am not sure how to handle that PR.  Comments from others?

iparask:
> but then again, I am sure @mturilli will not like that this changes the API (even with deprecation warning). 

Please check issue #481, it was opened by @mturilli :)

>Also, it misses the target a bit: RP [uses](https://github.com/radical-cybertools/radical.pilot/blob/devel/src/radical/pilot/compute_unit_description.py#L20) `cpu_threads`, not `cpu_threads_per_process` for example.

Oops I missed that. It's okay I can change it quickly.

andre-merzky:
> Please check issue #481, it was opened by @mturilli :)

Ha, then you have a case! :-)  Let's wait for him to comment anyway though, please.

iparask:
Pylint fails because I removed a py file for coverage.

mturilli (commented):
Thanks, this looks good and yes, seeing the level of adoption we now have, any change in API should be backwards compatible for several months if not years. Did I understand clearly that this PR generates only a warning if we don't use the new API? Why the PR deleted three tests?

mturilli:
Do we know why we get: ```************* Module tests/sitecustomize.py

tests/sitecustomize.py:1:0: F0001: No module named tests/sitecustomize.py (fatal)```

iparask:
Pylint check the diff between the PR and base branches. It cannot find it and reports it as an error.

mturilli:
Thanks. Do you see a way to fix that?

iparask:
> Did I understand clearly that this PR generates only a warning if we don't use the new API? 

Yes, it generates a Deprecation warning the first time to avoid a large number of warnings.

> Why the PR deleted three tests?

It updates `test_task.py` and removes two files that have to do with executing tests on Jenkins, since we do not use it anymore.

iparask:
> Thanks. Do you see a way to fix that?

I am not sure how to fix it. I'll check if there is a way to set pylint to exclude files that do not exist.

mturilli (approved):

iparask:
I can silence the error it produces, but I am afraid that it may cause pylint to pass in cases where something very wrong is happening.

Issue #493, opened 2020-10-22
Unit tests require more than 10 minutes to run on Travis.


iparask:
Hello @mturilli, @andre-merzky, @lee212 ,

I just realized that the unit tests require more than 10 minutes to run on Travis. I believe this has to do with the hypothesis package and it tests more than 100 test cases per test.

I was thinking of removing it and bring the unit test suite closer to RP.

Can you please let me know of your thoughts?

Thank you,
Giannis

andre-merzky:
It would certainly be good to reduce the test time - maybe that makes the tests also runnable on localhost again (I frequently do not manage to get EnTK tests to pass locally :-/).  But we also don't want to loose coverage in that reduction...  I also don't know enough about `hypothesis` and it's usage in the tests to argue one way or the other, I'm afraid...

mturilli:
As discussed, I think we need to understand pros/cons of using `hypothesis` and whether, if we decide to keep using it, we can speed it up or split it across multiple tests.

PR #494, opened 2020-10-26
Silence Integration and Issues tests

iparask:
Hello @mturilli, @andre-merzky, @lee212 
I would like to silence the integration and issues tests until issue #493 is fixed.

Can you offer me your thoughts? Thank you


andre-merzky:
Works for me, as long as we don't push re-enabling them off for too far.

iparask:
Yes

PR #496, opened 2020-10-28
Fix/unittests

iparask:
I am thinking of doing this incrementally. A lot of tests are still executing full blown examples. In this PR I refactored the unit tests for the following components:

- task
- stage
- pipeline
- wfprocessor
- Base_resourcemanager
- Base_taskmanager
- states
- general module import

Let me know of your thoughts and PR comments.

Pylint fails because of silenced tests and missing python files


src/radical/entk/appman/wfprocessor.py
```diff
@@ -299,6 +301,8 @@ def _update_dequeued_task(self, deq_task):
 
                 # Next search across all stages of a matching
                 # pipelines
+                stage = None
```
andre-merzky, line 304: That should never happen, and I would suggest to replace this with `assert(pipe.stages)` if it did happen.  In that case we got an orphaned task which does not belong to any stage which we need to flag.  We could also add an `assert(stage)` after the next for loop to ensure that the stage is correctly identified.  In general (but out of scope for this PR), we should replace the search with a (dict) lookup, as that will really not scale well for large workflows...

tests/test_component/test_states.py
```diff
@@ -1,86 +1,93 @@
...
+                                            2: states.SCHEDULING,
+                                            3: states.SUSPENDED,
+                                            10: [states.DONE, states.FAILED,
+                                                states.CANCELED]
+                                            })
+
+
+    def test_stage_states(self):
+
+        self.assertEqual(states._stage_state_values, {states.INITIAL: 1,
```
andre-merzky: Why is `assertEqual(a, literal)` better / how is it different from `assert(a == literal)`?

andre-merzky (commented):
Great to see progress on the RE tests...

tests/test_component/test_states.py
iparask: The assert commands are ignored when python is executed with the -O flag as they are meant to be used mainly for debugging. See [here](https://docs.python.org/3/reference/simple_stmts.html#the-assert-statement). Also, the methods offered by `TestCase` are supposed to provide better messages when a test fails ([here](https://docs.python.org/3/library/unittest.html#unittest.TestCase.assertEqual)).

Does this help?

src/radical/entk/appman/wfprocessor.py
iparask: I am perfectly fine with changing it. I did it due to this message from pylint:
```
src/radical/entk/appman/wfprocessor.py:350:19: W0631: Using possibly undefined loop variable 'stage' (undefined-loop-variable)
```

To me it makes sense. The `if` in that line ([here](https://github.com/radical-cybertools/radical.entk/blob/fix/unittests/src/radical/entk/appman/wfprocessor.py#L350)) makes the assumption that the for loop in [line 306](https://github.com/radical-cybertools/radical.entk/blob/fix/unittests/src/radical/entk/appman/wfprocessor.py#L306) finds a stage or it will return the last stage.

> We could also add an assert(stage) after the next for loop to ensure that the stage is correctly identified.

This will not fix the warning from pylint. Also, I do not thing it will work. Even if there is a orphan task, the last stage will populate the `stage` variable since that is how the for loop ended.

I am open to suggestions.

tests/test_component/test_states.py
andre-merzky: Not sure if we would ever run tests with `-O`, but the better message makes sense, thanks.

src/radical/entk/appman/wfprocessor.py
andre-merzky: Yeah, please do add the asserts.  As for the warning: it's useless in this case, but adding an exception is also not useful, so let's stick with it.

iparask: Done

andre-merzky (approved):
LGTM - thanks for bearing with me :-)

PR #497, opened 2020-10-31
Fix warning regarding ABC import from collections

### LINKED ISSUE #495: Collections ABCs are deprecated.
I get the following warning with EnTK:
```
../../../opt/anaconda3/envs/rct_test/lib/python3.8/site-packages/radical/entk/pipeline/pipeline.py:6
  /Users/iparask/opt/anaconda3/envs/rct_test/lib/python3.8/site-packages/radical/entk/pipeline/pipeline.py:6: DeprecationWarning: Using or importing the ABCs from 'collections' instead of from 'collections.abc' is deprecated since Python 3.3, and in 3.9 it will stop working
    from collections import Iterable
```
Although it is not important now, we should take care before it is too late.

tirkarthi:
Fixes #495

### DIFF
--- a/src/radical/entk/pipeline/pipeline.py
+++ b/src/radical/entk/pipeline/pipeline.py
@@ -3,7 +3,7 @@ from radical.entk.exceptions import *
 from radical.entk.stage.stage import Stage
 import threading
 from radical.entk import states
-from collections import Iterable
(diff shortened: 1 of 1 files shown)


iparask (approved):
Looks good to me. Thank you!

PR #498, opened 2020-11-02
Linting and fixing task path in appmanager

iparask:
I linted some of the files for bring them in a state where travis does not fail because of that!

Thank you


iparask:
This also moves where the path of a task is set in the appmanager.

src/radical/entk/appman/appmanager.py
```diff
@@ -835,6 +835,12 @@ def _update_task(self, msg, reply_to, corr_id, mq_channel, method_frame):
                         self._logger.debug(('Found task %s in state (%s)'
                             ' changing to %s ==') %
                                 (task.uid, task.state, completed_task.state))
+
+                        if completed_task.path:
+                            task.path = str(completed_task.path)
+                            self._logger.debug('Task %s path set to%s' %
+                                               (task.uid, task.path))
```
andre-merzky, line 842: Maybe a space between `to` and `%s`?

src/radical/entk/execman/mock/resource_manager.py
```diff
@@ -29,7 +29,7 @@ class ResourceManager(Base_ResourceManager):
 
     # --------------------------------------------------------------------------
     #
-    def __init__(self, resource_desc, sid, rts_config={}):
+    def __init__(self, resource_desc, sid, rts_config):
```
andre-merzky, line 33: Is the intent of that change to get in sync with the other RM implementations?

andre-merzky (approved):
LGTM - I have a cosmetic comment only.

src/radical/entk/execman/mock/resource_manager.py
iparask: RM? I am not sure what that is. I did the change because of this message from pylint:
```
 W0102: Dangerous default value {} as argument (dangerous-default-value)
```

src/radical/entk/appman/appmanager.py
iparask: Fixed

src/radical/entk/execman/mock/resource_manager.py
andre-merzky: `RM` - the ResourceManager whose constuctor signature you changed.

The fix for that linting error is:
```py
def __init__(self, resource_desc, sid, rts_config=None):
    if rts_config is None:
        rts_config = dict()
```

The reason for that is some more python by-value / by-reference / data localization stupidity (really, it's braindead).  The change you did changes the signature, which you may want to avoid.  The above does not change the signature, but delays the problematic default value assignment.

iparask: It's already happening in the base class: https://github.com/radical-cybertools/radical.entk/blob/fix/file_linting/src/radical/entk/execman/base/resource_manager.py#L34

andre-merzky: I really don't know what the signature *should* be - just saying that the syntactic-neutral change for the linting error would be a bit different.  Anyway, happy to leave that choice to you.

PR #499, opened 2020-11-02
Additional unit tests

### LINKED ISSUE #436: pytest never finishes when running locally
This is on
```sh
$ rs

  python               : 3.7.5
  pythonpath           :
  virtualenv           : /home/merzky/radical/radical.entk.devel/ve3

  radical.entk         : 1.0.1-v1.0.1-2-g92797c93@feature-early_uid
  radical.pilot        : 1.1.1
  radical.saga         : 1.1.2
  radical.utils        : 1.1.1
```

### LINKED ISSUE #490: Increase coverage to at least 40%
Unit tests should cover at least 40% of the code based until the November release

### LINKED ISSUE #493: Unit tests require more than 10 minutes to run on Travis.

iparask:
Added more unit tests. Unit tests for `appmanager` and `prof_utils` are missing. They will be included in the last and third stint.

This PR resolves #493 , resolves #490, resolves #436.


andre-merzky (approved):
LGTM.  I am not sure why the introduction of `resolved_path` is needed - code looked valid and readable before?  But it looks correct.

iparask:
> LGTM. I am not sure why the introduction of `resolved_path` is needed - code looked valid and readable before? But it looks correct.

The code was doing:
```
            for path in task.link_input_data:

                 path = resolve_placeholders(path, placeholders)
```

Although valid and readable, if there is something wrong with `path` it is not easy to understand if it is because of the loop or the function call. That is the reason for the change

andre-merzky:
> Although valid and readable, if there is something wrong with `path` it is not easy to understand if it is because of the loop or the function call. That is the reason for the change

You will find that in many places in the code.  Reason is that Python does not allow to take some types by-references, but always by-value.  One could for example write:
```py
for data in input:
    cleanup(data)
    compute(data)
    fout.write(data)
```

if `data` is a dict - but if `data` is a string or int, that would not work.  I'll withhold my opinion on that language feature, but that's why we have the explicit version in many places:
for data in input:
    data = cleanup(data)
    data = compute(data)
    fout.write(data)
```

No worries about this PR, that looks good - just saying that in var reuse is ok also ok because of the by-value thingie...

**Comment by andre-merzky** [comment]
Point in case:
```py
$ python3
Python 3.8.6 (default, Sep 25 2020, 09:36:53)
>>> input = ['foo', 1, ['bar']]
>>> def cleanup(data):
...     data *= 2
...
>>> for data in input:
...     cleanup(data)
...     print(data)
...
foo
1
['bar', 'bar']
```

iparask:
Either way, I reverted the change to make sure we are consistent!

iparask:
@andre-merzky, would you mind taking another look?

PR #507, opened 2020-11-18
Feature/id mapping

### LINKED ISSUE #500: EnTK and runtime id mapping
EnTK should provide a mapping between its task ids and runtimes ids. Essentially, when the workflow is written as a json file. Currently, a workflow looks like:
```
'workflows': [{'pipes': [{'name': 'pipe.0000',
                          'stages': [{'name': 'stage.0000',
                                      'state_history': ['SCHEDULING'],
                                      'tasks': [{'name': 'task.0000',
                                                 'state': 'DESCRIBED',
                                                 'state_history': ['DESCRIBED'],
                                                 'uid': 'task.0000'}],
```

the feature ask for:

```
'workflows': [{'pipes': [{'name': 'pipe.0000',
                          'stages': [{'name': 'stage.0000',
                                      'state_history': ['SCHEDULING'],
                                      'tasks': [{'name': 'task.0000',
                                                 'state': 'DESCRIBED',
                                                 'state_history': ['DESCRIBED'],
                                                 'uid': 'task.0000',
                                                 'rts_id': 'uid.0000'}],
```

iparask:
This PR closes #500.

The `rst_uid` of a task is now transmitted through the RMQ messages and is included in the json file that has the session. Although the field existed, it was not included in the task's dictionary version and as a result never picked up from the Appmanager.

In addition, I added a deprecation message for the `write_workflow` function that is offered by EnTK. Both function do the same thing in a slightly different way. There is no need for both of them.


src/radical/entk/task/task.py
```diff
@@ -548,6 +549,17 @@ def state_history(self):
 
         return self._state_history
 
+    @property
+    def rts_uid(self):
+        '''
+        Unique RTD ID of the current task
```
andre-merzky, line 555: RTD?

src/radical/entk/appman/appmanager.py
```diff
@@ -842,6 +842,11 @@ def _update_task(self, msg, reply_to, corr_id, mq_channel, method_frame):
                             self._logger.debug('Task %s path set to %s' %
                                                (task.uid, task.path))
 
+                        if completed_task.rts_uid:
```
andre-merzky, line 845: Should we maybe set this for running tasks already?  Otherwise the RTS uid will not be available for tasks which did not complete, for some reason?  Probably not much of a problem though...

src/radical/entk/appman/config.json
```diff
@@ -7,7 +7,7 @@
     "reattempts"      : 3,
     "resubmit_failed" : false,
     "autoterminate"   : true,
-    "write_workflow"  : false,
+    "write_workflow"  : true,
```
andre-merzky, line 10: Why was that changed?  I mean, I don't mind, just wonder about the reason...

andre-merzky (commented):
Left some comments, but this looks mostly good to me.

src/radical/entk/task/task.py
iparask: haha this is a typo!

src/radical/entk/appman/appmanager.py
iparask: I think it is set every time the app manager gets an update for a task and not only when they are completed. The method is called `_update_task` and the variable `completed_task`.

I will check and let you know

src/radical/entk/appman/config.json
iparask: No reason, I set up to test things and I think it just got committed. I will revert it. There is no reason to write a file every time EnTK is run.

src/radical/entk/appman/appmanager.py
iparask: The function that sets information like `path` and `rts_uid` for a task is called `create_task_from_cu` which is called in a unit call back ([here](https://github.com/radical-cybertools/radical.entk/blob/devel/src/radical/entk/execman/rp/task_manager.py#L249)). The callback only creates a task from a unit when the unit has finished executing. Every other task state update the Appmanager receives comes from somewhere else.

> Otherwise the RTS uid will not be available for tasks which did not complete, for some reason?

That is true. I do not think we should care about this case.

andre-merzky: If that implies structural changes to the code, then it's out of scope for this PR indeed.  Thanks for checking.

andre-merzky (approved):
LGTM, thanks.

PR #508, opened 2020-11-19
Unit tests for methods that use threads and processes

iparask:
This PR introduces unit tests for methods that are launching threads and processes. The tests are expanded for the base task manager, RP task manager and workflow processor. Unfortunately, I could introduce any additional tests for the application manager as most of its methods can be tested through integration tests. 

Any suggestions, changes or telling I am wrong is desirable!


tests/test_component/test_tmgr_base.py
```diff
@@ -9,16 +9,29 @@
...
 except ImportError:
     from unittest import mock
 
 
+# ------------------------------------------------------------------------------
+#
+def _tmgr_side_effect(event):
+
+    while not event.is_set():
+        continue
```
andre-merzky, line 28: you may want to add a small sleep (`0.01 or so`)here as not to have a busy wait loop ...?

iparask: Thank you!

tests/test_component/test_tmgr_base.py
```diff
@@ -114,3 +127,161 @@ def _sync_side_effect(log_entry, uid, state, msg):
...
+
+        rmq_params = mock.MagicMock(spec=ConnectionParameters)
+        rmgr = mock.MagicMock(spec=Base_ResourceManager)
+        tmgr = Tmgr('test_tmgr', ['pending_queues'], ['completed_queues'], 
+                     rmgr, rmq_params, 'test_rts')
+
+        def _tmgr_side_effect(amount):
+            time.sleep(amount)
+
+        tmgr._tmgr_process = mt.Thread(target=_tmgr_side_effect,
```
mturilli, line 181: I am missing something that is probably obvious but should we pass an `amount`?

mturilli (commented):
LGTM. Can you briefly summarize the logic of the tests? You mock the environments, create some threads and then test whether those thread were created? If so, under what conditions these tests could fail?

tests/test_component/test_tmgr_base.py
iparask: Yes, it is passed in the next line :) it is the `args` argument

iparask:
If the tested method creates a process the test checks whether the process actually exists in the operating system, by verifying the that the process ID actually exists. I want to do the same with threads, but I am not very sure yet how to check if a thread in the OS exists through python.
When the tested method terminates a process or thread, that is created by the test, terminated and verify that it is not there.

The methods that check the existence of a thread/ process are being extended to handle the case where the thread/ process do not exist.

The tests will fail in case those threads are not created correctly or not terminated correctly.

@mturilli, does this help?

mturilli:
Yes thanks.

mturilli (approved):

Issue #509, opened 2020-11-20
Connection reset while transferring stages


Weiming-Hu:
I'm testing on a small toy example. I have one executable that simply waits for 10 minutes. This workflow requires 1 core and 1 hour of resources. When the executable finishes waiting, EnTK got stuck and never made any progress.

Below error message can be found from the client-side logs:

```
1605825753.056 : radical.entk.task_manager.0000 : 9506  : 139943724709632 : DEBUG    : Unit unit.000000 in state DONE
1605825753.057 : radical.entk.task_manager.0000 : 9506  : 139943724709632 : INFO     : Transition task.0000 to EXECUTED
1605825753.057 : radical.entk.task_manager.0000 : 9506  : 139943724709632 : DEBUG    : task.0000 (EXECUTED) to sync with amgr
1605825753.058 : radical.entk.task_manager.0000 : 9506  : 139943724709632 : ERROR    : Transition task.0000 to state EXECUTED failed, error: (-1, "ConnectionResetError(104, 'Connection reset by peer')")
Traceback (most recent call last):
  File "/glade/u/home/wuh20/venv/lib/python3.7/site-packages/radical/entk/execman/base/task_manager.py", line 206, in _advance
    self._sync_with_master(obj, obj_type, channel, queue)
  File "/glade/u/home/wuh20/venv/lib/python3.7/site-packages/radical/entk/execman/base/task_manager.py", line 153, in _sync_with_master
    properties=pika.BasicProperties(correlation_id=corr_id))
  File "/glade/u/home/wuh20/venv/lib/python3.7/site-packages/pika/adapters/blocking_connection.py", line 2120, in basic_publish
    mandatory, immediate)
  File "/glade/u/home/wuh20/venv/lib/python3.7/site-packages/pika/adapters/blocking_connection.py", line 2207, in publish
    self._flush_output()
  File "/glade/u/home/wuh20/venv/lib/python3.7/site-packages/pika/adapters/blocking_connection.py", line 1292, in _flush_output
    *waiters)
  File "/glade/u/home/wuh20/venv/lib/python3.7/site-packages/pika/adapters/blocking_connection.py", line 477, in _flush_output
    result.reason_text)
pika.exceptions.ConnectionClosed: (-1, "ConnectionResetError(104, 'Connection reset by peer')")
1605825753.097 : radical.entk.task_manager.0000 : 9506  : 139943724709632 : DEBUG    : task.0000 (DESCRIBED) to sync with amgr
[...]

PR #513, opened 2020-12-02
Fix/issue 509

iparask:
This is to take care of cases like issue #509. I still to find a proper fix for that one.

### DIFF
--- a/src/radical/entk/appman/appmanager.py
+++ b/src/radical/entk/appman/appmanager.py
@@ -22,6 +22,8 @@ from ..utils       import write_workflows
 
 from .wfprocessor  import WFprocessor
 
+pika.connection.Parameters.DEFAULT_HEARTBEAT_INTERVAL = 0
+pika.connection.Parameters.DEFAULT_HEARTBEAT_TIMEOUT = 0
 
 # pylint: disable=protected-access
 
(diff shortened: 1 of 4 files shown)


src/radical/entk/execman/rp/task_manager.py
```diff
@@ -313,7 +314,9 @@ def unit_state_cb(unit, state):
 
                     self._advance(task, 'Task', states.SUBMITTING,
                                   mq_channel, '%s-tmgr-to-sync' % self._sid)
-
+                    self._log.debug('Connection State %s channel state %s'
+                                        % (mq_connection.is_open,
+                                           mq_channel.is_open))
```
andre-merzky, line 319: I would suggest to disable that message - it's on a hot code path.  OTOH, it's `DEBUG` level, and we should not encourage to run `DEBUG` as default level either.  Hmm, so probably ok... :-)

andre-merzky (approved):
LGTM

src/radical/entk/appman/appmanager.py
```diff
@@ -22,6 +22,8 @@
 
 from .wfprocessor  import WFprocessor
 
+pika.connection.Parameters.DEFAULT_HEARTBEAT_INTERVAL = 0
+pika.connection.Parameters.DEFAULT_HEARTBEAT_TIMEOUT = 0
```
mturilli, line 27: Do we want that to be the default behavior for EnTK and RBMQ? Do we have issues only with one user and one machine or is the heartbeat_interval going to be a more general issue? If the former, then should we keep the default as it has been so far and enable setting the interval/timeout to a specific value via env variables?

iparask:
Closing this PR as there is an easy workaround

Issue #517, opened 2020-12-07
AppMan does not fail after max attempts


iparask:
The application manager restarts the workflow processor, task manager, and synchronization thread if they fail. It allows though up to 3, as default, component restart regardless of the combination. If one of these components fails and causes the fourth failure, the application manager falls into an infinite loop.

This is observable by the latest unit test pushed in the `fix/unittests` branch. The error produced by the test is the following, after adding a timeout decorator:
```
        try:
            appman._run_workflow()
        except timeout_decorator.timeout_decorator.TimeoutError as e:
            print(e)
>           self.assertLess(appman._cur_attempt, 3)
E           AssertionError: 4 not less than 3

tests/test_component/test_amgr.py:187: AssertionError
----------------------------------------------------------------------------------------------- Captured stdout call -----------------------------------------------------------------------------------------------
'Timed Out'
================================================================================================= warnings summary =================================================================================================
../../../opt/anaconda3/envs/rct_test/lib/python3.8/site-packages/radical/utils/atfork/stdlib_fixer.py:77
  /Users/iparask/opt/anaconda3/envs/rct_test/lib/python3.8/site-packages/radical/utils/atfork/stdlib_fixer.py:77: UserWarning: logging handlers already registered.
    warnings.warn('logging handlers already registered.')

-- Docs: https://docs.pytest.org/en/stable/warnings.html
============================================================================================= short test summary info ==============================================================================================
FAILED tests/test_component/test_amgr.py::TestBase::test_run_workflow - AssertionError: 4 not less than 3
```

iparask:
Resolved

PR #518, opened 2020-12-07
AppManager raises error when there are multiple reattempts

iparask:
The AppManager now raises an error when it tries to restart components more than the maximum times set at the beginning of the execution.

This closes issue #517


mturilli (approved):
LGTM. Well spotted!

PR #519, opened 2020-12-08
Small refactoring changes.

iparask:
Exception chaining is done and flake8 and pylint should  be happy,


iparask:
I am not sure why git diff does not work. I would like to see the git message locally and then fix it. Currently, the code passes all checks.

lee212 (approved):
Looks good to me. thanks for the PR 👍

PR #522, opened 2020-12-11
Fix/integration tests

iparask:
Introducing integration tests and trying to setup a testing framework. I'll keep it as a draft until it is ready for review.


iparask:
This adds a single integration test; simultaneously, it adds linting tests that check only the changed python files that are not removed.

lee212 (approved):
LGTM, thanks for the PR 👍

PR #525, opened 2020-12-11
Moving warning filter to the whole file.

### LINKED ISSUE #523: EnTK deprecated warning is printed hundreds of times
@iparask a user reported an unintended consequence of the API deprecation message: it prints hundreds of times. We should be sure we print that only once.

iparask:
I moved the warning filter to the whole file. It will make sure that the deprecation message will be printed once, regardless how many times a task is created. I am not sure how to create a test other than running something like:

```python
from radical.entk import Task
for _ in range(100):
    t = Task()
    t.cpu_reqs = {'processes': 16, 'process_type': 'MPI', 'threads_per_process': 1, 'thread_type': 'OpenMP'}
for _ in range(100):
    t = Task()
    t.gpu_reqs = {'processes': 16, 'process_type': 'MPI', 'threads_per_process': 1, 'thread_type': 'OpenMP'}
```

and observing the output.

This fixes #523.

### DIFF
--- a/src/radical/entk/task.py
+++ b/src/radical/entk/task.py
@@ -7,6 +7,9 @@ import radical.utils as ru
 from . import exceptions as ree
 from . import states     as res
 
+import warnings
+warnings.simplefilter(action="once", category=DeprecationWarning, lineno=707)
+warnings.simplefilter(action="once", category=DeprecationWarning, lineno=764)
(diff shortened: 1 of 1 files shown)


mturilli (approved):
LGTM. Thanks!

PR #529, opened 2021-01-04
Added exception for any symbol apart from . in names

### LINKED ISSUE #392: `_` special character is not allowed in names of Pipeline, Stage, and Task; failing data sharing
To exchange or share data between tasks, e.g. `copy_input_data`, PST names are referenced but the underscore `_` is used for parsing. The errors, otherwise, is seen like, if names contain this character:

```
p = Pipeline()
p.name = 'simple_mdff'
...
EnTKError: Value for attribute placeholder of object task incorrect.
 Expected value ('$Pipeline_(pipeline_name)_Stage_(stage_name)_Task_(task_name) 
or $SHARED',),
but got ['$Pipeline', 'simple', 'mdff', 'Stage', 'Generating a simulated density map', 
'Task', 'Starting to load the target PDB'].
``` 

pipeline name `simple_mdff` was accidentally sliced by the delimiter `_`. 

A possible solution would be:
- encoding names by base64 as it contains three special characters only `+, /,  =`
- prohibit using `_` in names

https://github.com/radical-cybertools/radical.entk/blob/d249c2c95db5987c42ed84a52e796e00ac8f4e2a/src/radical/entk/execman/rp/task_processor.py#L59

iparask:
This PR fixes #392.

It raises an exception if any symbol other than '.' is used in the name of a task, stage, or pipeline.

### DIFF
--- a/src/radical/entk/pipeline.py
+++ b/src/radical/entk/pipeline.py
@@ -6,9 +6,10 @@ import threading
 
 import radical.utils as ru
 
-from .exceptions import TypeError, ValueError, MissingError, EnTKError
-from .stage      import Stage
-from .           import states
+from string import punctuation
+from .      import exceptions as ree
+from .stage import Stage
+from .      import states
 
 
 class Pipeline(object):
@@ -156,20 +157,21 @@ class Pipeline(object):
 
     @name.setter
     def name(self, value):
-        if isinstance(value, str):
-            if ',' in value:
-                raise ValueError(obj=self._uid,
-                                attribute='name',
-                                actual_value=value,
-                                expected_value="Using ',' in an object's " +
-                                               "name will corrupt the " +
-                                               "profiling and internal " +
-                                               "mapping tables")
-            else:
-                self._name = value
 
-        else:
-            raise TypeError(expected_type=str, actual_type=type(value))
+        invalid_symbols = punctuation.replace('.','')
+        if not isinstance(value, str):
+            raise ree.TypeError(expected_type=str,
+                                actual_type=type(value))
+
+        if any(symbol in value for symbol in invalid_symbols):
+            raise ree.ValueError(obj=self._uid,
+                                 attribute='name',
+                                 actual_value=value,
+                                 expected_value="Valid object names can " +
+                                 "contains letters, numbers and '.'. Any "
+                                 "other character is not allowed")
+
+        self._name = value
 
     @stages.setter
     def stages(self, value):
@@ -190,12 +192,12 @@ class Pipeline(object):
                 if self._state != states.SUSPENDED:
                     self._state_history.append(value)
             else:
-                raise ValueError(obj=self._uid,
+                raise ree.ValueError(obj=self._uid,
                                  attribute='state',
                                  expected_value=list(states._pipeline_state_values.keys()),
                                  actual_value=value)
         else:
-            raise TypeError(expected_type=str, actual_type=type(value))
+            raise ree.TypeError(expected_type=str, actual_type=type(value))
 
 
     # --------------------------------------------------------------------------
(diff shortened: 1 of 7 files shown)


iparask:
@iparask: What is the performance overhead?

src/radical/entk/pipeline.py
```diff
@@ -156,20 +157,21 @@ def state_history(self):
...
+        if not isinstance(value, str):
+            raise ree.TypeError(expected_type=str,
+                                actual_type=type(value))
+
+        if any(symbol in value for symbol in invalid_symbols):
+            raise ree.ValueError(obj=self._uid,
+                                 attribute='name',
+                                 actual_value=value,
+                                 expected_value="Valid object names can " +
+                                 "contains letters, numbers and '.'. Any "
```
mturilli, line 171: contain?

src/radical/entk/pipeline.py
```diff
@@ -254,19 +256,32 @@ def from_dict(self, d):
...
+                    raise ree.TypeError(expected_type=str,
+                                        actual_type=type(d['name']))
+
+                if any(symbol in d['name'] for symbol in invalid_symbols):
+                    raise ree.ValueError(obj=self._uid,
+                                        attribute='name',
+                                        actual_value=d['name'],
+                                        expected_value="Valid object names can " +
+                                        "contains letters, numbers and '.'. Any "
+                                        "other character is not allowed")
```
mturilli, line 270: Would it be worth to store these messages in a dedicated data structure? See that you repeated the same message twice, and now we have to fix the same error in both instances of the same message. More in general, should we template exceptions? Repeating `attribute` and other parameters all the time might be undesirable.

src/radical/entk/stage.py
```diff
@@ -166,16 +167,19 @@ def func_post():
...
+        if not isinstance(value, str):
+            raise ree.TypeError(expected_type=str,
+                                actual_type=type(value))
+
+        if any(symbol in value for symbol in invalid_symbols):
+            raise ree.ValueError(obj=self._uid,
+                                 attribute='name',
+                                 actual_value=value,
+                                 expected_value="Valid object names can " +
+                                 "contains letters, numbers and '.'. Any "
```
mturilli, line 180: See the comments made before about the typo, templating and messages

src/radical/entk/stage.py
```diff
@@ -271,6 +275,18 @@ def from_dict(self, d):
...
+                if not isinstance(d['name'], str):
+                    raise ree.TypeError(expected_type=str,
+                                        actual_type=type(d['name']))
+
+                if any(symbol in d['name'] for symbol in invalid_symbols):
+                    raise ree.ValueError(obj=self._uid,
+                                        attribute='name',
+                                        actual_value=d['name'],
+                                        expected_value="Valid object names can " +
+                                        "contains letters, numbers and '.'. Any "
```
mturilli, line 288: Same as the previous comment

src/radical/entk/task.py
```diff
@@ -593,17 +595,18 @@ def rts_uid(self, value):
...
 
-        if ',' in value:
+        if any(symbol in value for symbol in invalid_symbols):
             raise ree.ValueError(obj=self._uid,
                                  attribute='name',
                                  actual_value=value,
-                                 expected_value="Using ',' in an object's name"
-                                 "will corrupt internal mapping tables")
+                                 expected_value="Valid object names can " +
+                                 "contains letters, numbers and '.'. Any "
```
mturilli, line 608: Same as the previous comment

mturilli (commented):
Left some comments.

src/radical/entk/pipeline.py
iparask: Done

src/radical/entk/pipeline.py
iparask: I created a `constants` file to keep constants in general. I placed the message there. Some exceptions are already templated, but I am not a fan because it confuses me a bit. I can do it if you think it is important.

src/radical/entk/stage.py
iparask: Done

src/radical/entk/pipeline.py
andre-merzky: This should only be done for messages we use a lot IMHO.

mturilli: This message is used ~8 times and each instance had the same typo. Happy for you to have the last word.

mturilli: @iparask, about the templating, happy to follow your preference as you are the main developer of EnTK. I would probably stress the need of consistency so that we do not end up with a mix of (non)templated exceptions?

mturilli (approved):

PR #530, opened 2021-01-04
pypi version check

### LINKED ISSUE #476: Version update message for older stacks
To encourage using the latest version of RCT, it is useful if a version check is enabled with a message to update.

iparask:
This PR checks the version of the installed EnTK and the latest version on pypi and produces a warning message if the pypi version is more recent:

```
WARNING: You are using radical.entk version 1.5.9, however version 1.5.10 is available.
```

It will make sense two releases from now.

Fixes #476

### DIFF
--- a/src/radical/entk/__init__.py
+++ b/src/radical/entk/__init__.py
@@ -11,14 +11,26 @@ from .appman.appmanager import AppManager
 
 # ------------------------------------------------------------------------------
 #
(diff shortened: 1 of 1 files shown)


iparask:
ping @lee212, @andre-merzky

mturilli (approved):
LGTM.

PR #535, opened 2021-01-12
Examples sync to devel

iparask:
This PR fixes the examples to use the new naming convention and demotes the exception to a warning. The warning will appear for all names always.

It will probably annoy our users, but I hope they do the change instead of asking us to silence it.

### DIFF
--- a/src/radical/entk/constants.py
+++ b/src/radical/entk/constants.py
@@ -3,4 +3,4 @@ __copyright__ = 'Copyright 2014-2020, http://radical.rutgers.edu'
 __license__   = 'MIT'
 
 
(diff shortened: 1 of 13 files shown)


src/radical/entk/constants.py
```diff
@@ -3,4 +3,4 @@
 __license__   = 'MIT'
 
 
-NAME_MESSAGE = "Valid object names can contain letters, numbers and '.'. Any other character is not allowed"
+NAME_MESSAGE = "Valid object names can contain letters, numbers and '.'. Any other character will not be allowed in version 1.5.10"
```
mturilli, line 6: "Valid object names can contain letters, numbers and the symbol '.'. Any other character will not be allowed in version >= 1.5.10"?

mturilli (approved):
Left a comment. Once addressed, this is ready for me to go.

PR #538, opened 2021-01-25
EnTK components failure resilience

iparask:
This PR introduces full support from failing EnTK components. The failed component is restarted and any task that is not executed is resubmitted.


iparask:
I would like to go through a review process before merging with the cu to task refactoring. Let me know what you think

iparask:
@andre-merzky, @mturilli this PR ready for review from my side. Let me know what you think.

PR #539, opened 2021-01-27
Fix/issue 509

iparask:
This PR fixes the connection closed exception we were getting for long-running tasks. When the connection or the channel is closed, we receive an exception from pika. We then handle that exception by creating a new connection and channel and send the message again.

Unfortunately, we cannot know the actual RMQ server state for a connection or channel through their `is_open` or `is_closed` methods.

Although there is a replication in the messaging code, we avoid the overhead of creating a new connection and channel for every message.

### DIFF
--- a/.codecov.yml
+++ b/.codecov.yml
@@ -6,3 +6,4 @@ coverage:
         threshold: 1%
         paths: 
           - "src"
+    patch: off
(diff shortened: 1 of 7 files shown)


andre-merzky (approved):
LGTM.

mturilli (approved):
Looks good to me. I would not have renamed `channel` but if you prefer that I am fine with it.

iparask:
> I would not have renamed `channel` but if you prefer that I am fine with it.

I did it because I was getting some error somewhere, but I do not remember right now what it was.

Thank you both for reviewing!

PR #567, opened 2021-04-08
First attempt of verifying a task can fit a resource

iparask:
Hello @andre-merzky, @mtitov,

This is how I think EnTK can filter tasks that are larger than the resources it can have. What do you think? If you agree I will continue with this feature and include its tests as well.

### DIFF
--- a/src/radical/entk/execman/rp/task_manager.py
+++ b/src/radical/entk/execman/rp/task_manager.py
@@ -61,6 +61,8 @@ class TaskManager(Base_TaskManager):
         self._log.info('Created task manager object: %s', self._uid)
         self._prof.prof('tmgr_create', uid=self._uid)
         self._rp_tmgr = None
+        self._total_res = {'cores': 0,
+                           'gpus': 0}
 
 
     # --------------------------------------------------------------------------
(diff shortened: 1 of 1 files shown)


src/radical/entk/execman/rp/task_manager.py
```diff
@@ -282,6 +286,17 @@ def load_placeholder(task):
                                                             {'path': task.path,
                                                              'uid': task.uid}
 
+        # ----------------------------------------------------------------------
+        def check_resource_reqs(task):
+
+            cpu_reqs = task.cpu_reqs['cpu_processes'] * task.cpu_reqs['cpu_threads']
+            gpu_reqs = task.gpu_reqs['gpu_processes'] * task.gpu_reqs['gpu_threads']
```
andre-merzky, line 293: RP currently interprets `gpu_processes` as `gpus_per_process`, and ignores `gpu_threads`.  So the above should ikely be:
```py
cpu_reqs = task.cpu_reqs['cpu_processes'] * task.cpu_reqs['cpu_threads']
gpu_reqs = task.cpu_reqs['cpu_processes'] * task.gpu_reqs['gpu_processes']
```

andre-merzky (approved):
Looks good in general - one comment about how RP interprets the task description (at the moment)

src/radical/entk/execman/rp/task_manager.py
iparask: I'll fix this and move forward with the rest. Thank you!

PR #568, opened 2021-04-09
Using  RPs tags instead of tag

iparask:
I updated RE's integration with RP. Instead of `tag` for tagging tasks RE now uses `tags`.

Thank you

### DIFF
--- a/src/radical/entk/execman/rp/task_processor.py
+++ b/src/radical/entk/execman/rp/task_processor.py
@@ -473,9 +473,9 @@ def create_td_from_task(task, placeholders, task_hash_table, pkl_path, sid,
         td.stage_on_error = task.stage_on_error
 
         if task.parent_pipeline['uid']:
-            td.tag = resolve_tags(task=task,
+            td.tags = {'colocate': resolve_tags(task=task,
                                   parent_pipeline_name=task.parent_pipeline['uid'],
-                                  placeholders=placeholders)
+                                  placeholders=placeholders)}
 
         td.cpu_processes    = task.cpu_reqs['cpu_processes']
         td.cpu_threads      = task.cpu_reqs['cpu_threads']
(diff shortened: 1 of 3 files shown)


andre-merzky (approved):
LGTM!

src/radical/entk/execman/rp/task_processor.py
```diff
@@ -473,9 +473,9 @@ def create_td_from_task(task, placeholders, task_hash_table, pkl_path, sid,
         td.stage_on_error = task.stage_on_error
 
         if task.parent_pipeline['uid']:
-            td.tag = resolve_tags(task=task,
+            td.tags = {'colocate': resolve_tags(task=task,
```
mtitov, line 476: I'd like to propose that method `resolve_tags` returns the actual structure of `tags` attribute, since it is responsible to resolve it
```python
def resolve_tags(...
    ...
    return {'colocate': task.uid}
```

iparask: This is now resolved. @mtitov can you check if this is okay with you?

iparask:
I removed the test lines. I do not think the `if` is necessary. `placeholders` are defined in [line 265](https://github.com/radical-cybertools/radical.entk/blob/fix/deprecated_tags/src/radical/entk/execman/rp/task_manager.py#L265) when the task manager is instantiated. @mtitov, if you think it is necessary, I'll add it.

mtitov:
@iparask I don't have a strong opinion here, just proposed that since it was the case for a call within the test ([here](https://github.com/radical-cybertools/radical.entk/blob/7777dd94da5456925a304745a0a74fbe5a863f2f/tests/test_component/test_tproc_rp.py#L194)) which would trigger an error

mtitov (approved):
LGTM

PR #570, opened 2021-04-14
Update Task to Munch

iparask:
Hey @andre-merzky, this is the PR with the updates on the task description. I see the following error when I'm trying to validate a task in a python interpreter:

```python
>>> from radical.entk import Task
>>> a = Task()
>>> a.gpu_reqs = {'processes': 1, 'threads_per_process': 1, 'process_type':None, 'thread_type': 'OpenMP'}
>>> a.verify()
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/Users/iparask/opt/anaconda3/envs/rct_devel/lib/python3.9/site-packages/radical/utils/munch.py", line 294, in verify
    self._data[k] = self._verify_kvt(k, v, self._schema[k])
  File "/Users/iparask/opt/anaconda3/envs/rct_devel/lib/python3.9/site-packages/radical/utils/munch.py", line 276, in _verify_kvt
    if issubclass(t, Munch)   : return v.verify(t)
AttributeError: 'dict' object has no attribute 'verify'
>>>
```

I hope this helps.


andre-merzky:
TODO MTi: complete work on RU first before continuing that PR

mturilli:
> TODO MTi: complete work on RU first before continuing that PR

Do we have a PR on utils to link to this PR?

mtitov:
The idea was to extend `ru.Munch` functionality with method `_post_setter` (or `_pre_setter`), which can be overwritten in inherited classes (to have a checker per attribute assignment). Such enhancement would decrease the overall performance, but we didn't conduct corresponding experiments to estimate it.

This PR has work around it. And some [small] changes are still needed to finalize it -> WIP

setup.py
```diff
@@ -67,7 +67,7 @@ def get_version(_mod_root):
         if not src_root:
             src_root = '.'
 
-        with open(src_root + '/VERSION', 'r') as f:
+        with open(src_root + '/VERSION', 'r', encoding='utf-8') as f:
```
andre-merzky, line 70: There is an `ru.ru_open` which does the same to ensure we use the same encoding everywhere.

src/radical/entk/task.py
```diff
@@ -1,25 +1,110 @@
...
+
 import radical.utils as ru
 
-from string     import punctuation
 from .constants import NAME_MESSAGE
-
 from . import exceptions as ree
 from . import states     as res
 
-warnings.simplefilter(action="once", category=DeprecationWarning, lineno=728)
```
andre-merzky, line 14: off topic: we should at some point discuss when we will retire the deprecation warnings.

src/radical/entk/task.py
```diff
@@ -1,25 +1,110 @@
...
+        'thread_type'        : 'gpu_thread_type'
+    }
 
 
 # ------------------------------------------------------------------------------
 #
-class Task(object):
-    '''
+class Task(ru.Munch):
+    """
```
andre-merzky, line 107: nitpick: we actually use `'''` throughout most parts of the code.

src/radical/entk/task.py
```diff
@@ -29,1135 +114,229 @@ class Task(object):
...
-                                    actual_type=type(from_dict))
+        return not self == other
 
-            for k,v in from_dict.items():
-                self.__setattr__(k, v)
+    # --------------------------------------------------------------------------
+    #
+    def __hash__(self):
 
+        # FIXME: review having attribute "name" here
```
andre-merzky, line 226: agree - the UID itself should be sufficient.  Also, name might be unset (`None`) - what happens then, is that hashable?

src/radical/entk/task.py
```diff
@@ -29,1135 +114,229 @@ class Task(object):
...
+        t_elem = self['name']
 
-        p_elem = self.parent_pipeline.get('name')
-        s_elem = self.parent_stage.get('name')
-        t_elem = self.name
-
-        if not p_elem: p_elem = self.parent_pipeline['uid']
-        if not s_elem: s_elem = self.parent_stage['uid']
-        if not t_elem: t_elem = self.uid
+        if not p_elem: p_elem = self['parent_pipeline']['uid']
```
andre-merzky, line 299: Why did that change actually - shouldn't `Munch` keep this valid code?

src/radical/entk/task.py
```diff
@@ -29,1135 +114,229 @@ class Task(object):
...
+                                 attribute='state',
                                  expected_value=res.INITIAL,
-                                 actual_value=self._state)
+                                 actual_value=self['state'])
 
-        if not self._executable:
-            raise ree.MissingError(obj=self._uid,
+        if not self['executable']:
+            raise ree.MissingError(obj=self['uid'],
                                    missing_attribute='executable')
```
andre-merzky, line 341: Maybe add a call to `self.verify()`?

andre-merzky (changes requested):
Thanks all - this looks good to me overall.  The only actual complain I have is that we miss out on documentation.  We don't have a well-defined way to add documentation to `Munch` classes, but need to think of how to do that before we can merge this.  Let's open that can of worms on the next devel call.

setup.py
mtitov: but as I understand it is not the case for `setup.py` file (same for other packages)
if RU is not installed then installation will crash without trying to install its dependency

src/radical/entk/task.py
mtitov: actually following docstring convention it is recommended to use triple double quotes ([PEP257](https://www.python.org/dev/peps/pep-0257/) referred from [PEP8](https://www.python.org/dev/peps/pep-0008/#string-quotes))

> we actually use `'''` throughout most parts of the code.

not within EnTK ;) (almost all modules here use triple double quotes)

src/radical/entk/task.py
mtitov: made changes to use `uid` only

> Also, name might be unset (`None`) - what happens then, is that hashable?

hashable, and just return same value

src/radical/entk/task.py
mtitov: since it is internal calls, I wanted to decrease number of attribute-resolving calls: accessing items is faster than accessing attributes

src/radical/entk/task.py
```diff
@@ -29,1135 +114,227 @@ class Task(object):
...
-            if not isinstance(from_dict, dict):
-                raise ree.TypeError(expected_type=dict,
-                                    actual_type=type(from_dict))
+        return not self == other
 
-            for k,v in from_dict.items():
-                self.__setattr__(k, v)
+    # --------------------------------------------------------------------------
+    #
+    def __hash__(self):
```
eirrgang, line 223: This doesn't seem right. Instances are mutable, aren't they?

src/radical/entk/task.py
```diff
@@ -29,1135 +114,227 @@ class Task(object):
...
 
-        # Keep track of res attained
-        self._state_history = [res.INITIAL]
+        if not isinstance(other, Task):
+            return False
 
-        # Stage and pipeline this task belongs to
-        self._p_stage    = {'uid': None, 'name': None}
-        self._p_pipeline = {'uid': None, 'name': None}
+        return self['uid'] == other['uid']
```
eirrgang, line 213: This seems like a confusing/dangerous shortcut. Wouldn't it be better to explicitly compare `task1.uid == task2.uid`? If not, could you be clear about the new comparison behavior in the class documentation?

mtitov: This is the same, since here we introduce Task as a Munch-based class (== extended typed dictionary), and `task1.uid` is the same as `task1['uid']`, and since it is inside the class definition accessing items is faster than accessing attributes

src/radical/entk/task.py
eirrgang: Specifically, this implementation allows for the result of `__hash__` to change during the lifetime of the instance (if the 'uid' is set or reset), which is inconsistent with the Python data model, even if it seems like an edge case.

src/radical/entk/task.py
eirrgang: I understand. I meant that, since this is a new comparison, it makes more sense to continue to compare the uid fields, explicitly, rather than to introduce a novel shorthand by which `task1 == task2` based only on its `uid` field. At the very least, such novel behavior comes with a burden of clear documentation.

mtitov: don't see any contradiction.. this method was added to be in sync with `__hash__`, and just to make it explicit that `uid` has to be unique. Tasks instances are considered equals if the have the same `uid`. You always can compare certain fields explicitly, e.g. `t1.uid == t2.uid`, `t1.name == t2.name`, `t1.as_dict() == t2.as_dict()`

src/radical/entk/task.py
eirrgang: From https://docs.python.org/3/reference/datamodel.html#object.__hash__: "If a class defines mutable objects and implements an __eq__() method, it should not implement __hash__(), since the implementation of hashable collections requires that a key’s hash value is immutable (if the object’s hash value changes, it will be in the wrong hash bucket)."

Additionally, the hashability and related discussions regarding `dataclasses` and the various Mappings establish a strong pattern that should be clearly documented if rejected. See also https://docs.python.org/3/library/dataclasses.html?highlight=unsafe_hash

My suggestion is either
* make `uid` immutable if it is the basis of `__hash__`, and clearly document the comparison behaviors, or
* don't implement `__hash__`

I think it would be cleanest to refrain from adding either `__eq__` or `__hash__` with the current semantics.

src/radical/entk/task.py
eirrgang: > Tasks instances are considered equals if the have the same uid.

Does this logic exist prior to this PR? I'm sorry if I overlooked that. I thought `Task.__eq__` and `Task.__hash__` were new with this change.

eirrgang: > to make it explicit that uid has to be unique

It doesn't read that way to me. I read it more like "There may be multiple instances of Task that are assumed to refer to the same thing, but we have reasons to allow fields to be unpopulated or inconsistent, but we suppress the ability to check whether two instances are exact copies or not."

eirrgang: If you want to clarify the uniqueness requirement, I think documentation should come before technical enforcement. Technical enforcement could reasonably be considered beyond the scope of this PR, if there wasn't previously something in place.

If you do want to enforce the uniqueness, it should occur through the creation methods or manager objects.

One option would be a class variable storing a WeakValueDictionary mapping UIDs to instances, and either refuse to allow copies to exist or come up with a scheme to manage them. But the precedent so far is that it is not the responsibility of Task to guarantee the uniqueness of the uids of its instances (right?), and I think that is fine.

src/radical/entk/task.py
mtitov: @eirrgang Eric, thank you! That is helpful. We should be able to hash instances of Task since they are collected in `Stage` instance's attribute of type `set`, and thus `uid` will be set during initialization only.

src/radical/entk/task.py
mtitov: Original implementation does check uniqueness implicitly during the validation process ([here](https://github.com/radical-cybertools/radical.entk/blob/e2bc28a1ab5ac7197d9738b03bc3f3f8a0361ac1/src/radical/entk/task.py#L325))

I think what you describe is applicable if we would compare data structures, thus we say such objects are equal if all corresponding attributes are equal. In our case we check does two objects represent the same task, which is identified by ID (`uid`) and not by all set of attributes. EnTK (as well as RP) guarantees to create tasks with unique IDs within an application run.

andre-merzky (approved):
LGTM

PR #571, opened 2021-04-14
RP task processor adds default values to RP task cpu and gpu reqs

iparask:
RP's scheduler does not like `None` values for process and thread types. This PR makes sure that posix and openmp are used as the default values when RP is used as the RTS. Since RE can use any type of RTS, it did not make sense to me to make the `task` default values.


andre-merzky (approved):
LGTM.  RP will be able to accept `None` soon (that trickles up from a fix in RU), but the PR makes defaults explicit on the RE level, and that I think is good practice anyway.

PR #572, opened 2021-04-15
Trying to revert to codecov action after resetting the token

iparask:


iparask:
I am suggesting the following revert. I am not sure what is going on again with codecov. I hope it works later.

andre-merzky (approved):
Thank you!

PR #585, opened 2021-07-09
Fix/issue 581

lee212:
data staging by `.uid`, and the current branch is ready to merge.


andre-merzky (commented):
Do we need to update the documentation for this change?

lee212:
> Do we need to update the documentation for this change?

The current document is based on `.name` to locate files, so it will be okay with this change and I reviewed the current documentation and made a quick commit to reflect this change.  
https://github.com/radical-cybertools/radical.entk/pull/585/commits/e34903f48bfe0bd44d18b4b403a8d714621da6f7

lee212:
Note that the current PR does not check whether `.name` is unique or not. So if it is duplicate, the workflow will break to locate files.
It might have to address unique validation later in the naming.

andre-merzky (approved):
Thank you @lee212 !

Issue #589, opened 2021-07-20
Change default base path


lee212:
This is a feature request (from DeepDriveMD) to change the base path of logs/profiles from `pwd` to a user-defined location so that the session `re.session.xxx` would be placed on a different path when it's given.

For the reference, RP supports this, https://github.com/radical-cybertools/radical.pilot/blob/devel/src/radical/pilot/session.py#L107-L111,
```
>>> rp.Session(cfg={'base':'/tmp/test'})
```
The default base path is changed when a new path is given to the configuration argument, `cfg`, resulting in:
```
new session: [rp.session.login3.hrlee.018828.0003]                             \
...
>>> os.getcwd()
'/ccs/home/hrlee'
>>> os.listdir('/tmp/test')
['rp.session.login3.hrlee.018828.0003']
```

EnTK can apply the same approach, for example, `AppManager` can introduce a new argument `cfg` to take additional configuration from a user and change the default value.

PR #590, opened 2021-07-20
'base_path' param introduced in AppManager to change default directory of logger and profiler

lee212:
To address #589 , AppManager takes `config` param, python dictionary, to change default value such as base path for logger and profiler.

### DIFF
--- a/src/radical/entk/appman/appmanager.py
+++ b/src/radical/entk/appman/appmanager.py
@@ -76,7 +76,7 @@ class AppManager(object):
                  rmq_cleanup=None,
                  rts_config=None,
                  name=None,
-                 config={}):
+                 config=None):
 
         # Create a session for each EnTK script execution
         if name:
@@ -85,6 +85,9 @@ class AppManager(object):
         else:
             self._name = str()
             self._sid  = ru.generate_id('re.session', ru.ID_PRIVATE)
+
+        if config is None:
+            config = dict()
 
         self._read_config(config_path, hostname, port, username, password,
                           reattempts, resubmit_failed, autoterminate,
@@ -147,8 +150,8 @@ class AppManager(object):
 
         config = ru.read_json(os.path.join(config_path, 'config.json'))
 
-        # config_params (2nd dictionary) overwrites values if duplicated
-        config = { **config, **config_params }
+        # config_params overwrites values if duplicated
+        config.update(config_params)
 
         def _if(val1, val2):
             if val1 is not None: return val1
--- a/src/radical/entk/appman/appmanager.py
+++ b/src/radical/entk/appman/appmanager.py
@@ -154,7 +154,7 @@ class AppManager(object):
             if val1 is not None: return val1
             else               : return val2
 
-        self._path             = config.get('base_path', os.getcwd())
+        self._path             = str(config.get('base_path', os.getcwd()))
         self._hostname         = _if(hostname,        config['hostname'])
         self._port             = _if(port,            config['port'])
         self._username         = _if(username,        config['username'])
--- a/src/radical/entk/appman/appmanager.py
+++ b/src/radical/entk/appman/appmanager.py
@@ -58,6 +58,8 @@ class AppManager(object):
                           True/False} when RTS is RP
         :name:            Name of the Application. It should be unique between
                           executions. (default is randomly assigned)
+        :base_path:       base path of logger and profiler with a session id,
+                          None (default) will use a current working directory
     '''
 
     # --------------------------------------------------------------------------
@@ -76,7 +78,7 @@ class AppManager(object):
                  rmq_cleanup=None,
                  rts_config=None,
                  name=None,
-                 config=None):
+                 base_path=None):
 
         # Create a session for each EnTK script execution
         if name:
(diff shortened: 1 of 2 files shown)


src/radical/entk/appman/appmanager.py
```diff
@@ -75,7 +75,8 @@ def __init__(self,
                  rts=None,
                  rmq_cleanup=None,
                  rts_config=None,
-                 name=None):
+                 name=None,
+                 config={}):
```
andre-merzky, line 79: @lee212: I'm afraid you can't use complex types as default values in Python.  Consider the following script:
```py
#!/usr/bin/env python3

# ------------------------------------------------------------------------------
#
def test(cfg={}):
    if not cfg:
        print('no  cfg')
        cfg['foo'] = 'bar'
    else:
        print('has cfg: %s' % cfg)


# ------------------------------------------------------------------------------
#
if __name__ == '__main__':

    test()
    test()


# ------------------------------------------------------------------------------

```
which has the unexpected result:
```sh
$ ./t.py
no  cfg
has cfg: {'foo': 'bar'}
```

You will have to change that to:
```py
def test(cfg=None):
    if cfg is None:
        cfg = dict()
```

src/radical/entk/appman/appmanager.py
```diff
@@ -138,17 +140,21 @@ def __init__(self,
...
-                     write_workflow, rts, rmq_cleanup, rts_config):
+                     write_workflow, rts, rmq_cleanup, rts_config, config_params):
 
         if not config_path:
             config_path = os.path.dirname(os.path.abspath(__file__))
 
         config = ru.read_json(os.path.join(config_path, 'config.json'))
 
+        # config_params (2nd dictionary) overwrites values if duplicated
+        config = { **config, **config_params }
```
andre-merzky, line 151: Maybe `config.update(config_params)` would be more explicit and readable?

src/radical/entk/appman/appmanager.py
lee212: Thanks for the comment which makes sense. I applied this.

src/radical/entk/appman/appmanager.py
lee212: I agree, thanks for the comment.

src/radical/entk/appman/appmanager.py
```diff
@@ -75,7 +75,8 @@ def __init__(self,
                  rts=None,
                  rmq_cleanup=None,
                  rts_config=None,
-                 name=None):
+                 name=None,
+                 config=None):
```
mtitov, line 79: could the explicit name be used (`base_path`, `work_dir`, `session_dir`, etc.)? too many "config"s as inputs, and if to have such cumulative parameters, then all inputs should be reviewed, e.g., `rmq_settings` (which would include not just host and port, but ssl flag and vhost, and other related to RMQ), `app_settings` (which would include necessary `path`s, etc.), `rts_settings`.

lee212: I don't mind using the explicit name, `base_path` instead of `config`. It just doesn't make sense to me that AppManager carries out all the arguments which will grow over time.

mtitov: agree about growing over time, thus would propose to have some groupings
> e.g., `rmq_settings` (which would include not just host and port, but ssl flag and vhost, and other related to RMQ), `app_settings` (which would include necessary paths, etc.), `rts_settings`.

andre-merzky: @lee212 , @mtitov : I agree with both of you that a growing arg list is challenging.  However, please keep in mind that RE is supposed to have a very stable API, and we should avoid breaking changes to method signatures for the time being (adding `config` is not breaking, but renaming / re-grouping would be).  

The type checking work has some impact on the API, and the upcoming changes to `name` attribute and it's interpretation also.  Maybe we should collect those issues and this config discussion into a separate milestone which allows for a (minor) iteration of the API in one release instead of spreading over multiple releases?

There was also a discussion about a more radical (ha!) overhaul of the API - but I don't think there is sufficient consensus on the need for that right now, so I would leave that off the table.

@mturilli : any opinion?

mtitov: @andre-merzky that could be done transparently - keep `**kwargs` as input and check original parameters from there, while bringing up a new approach: 
```python
def __init__(self, config_path=None, app_settings=None, <other settings>, **kwargs):

    def _get_old_inputs(**kwargs):
        hostname = kwargs.get('hostname')
        ...
```
agree to keep it as a separate ticket/PR and just to set an explicit name for a new parameter for now

lee212: I also agree all the points made here and I don't want to break the discussion about managing API, but my reading is that we want to have `base_path` for this specific PR/ticket here instead of `config`. Because I'd like to make modifications accordingly.

lee212: @mtitov , @andre-merzky thanks for confirming. I made a commit to use an explicit parameter name.

andre-merzky (approved):
LGTM

andre-merzky:
@lee212 : please note that the tests are failing - please fix before merge.  Thanks

lee212:
> @lee212 : please note that the tests are failing - please fix before merge. Thanks

a310671 includes an updated list of arguments, test are not failing anymore.

Issue #591, opened 2021-07-20
Init fails when offline


mturilli:
Stack:
```
  radical.analytics    : 1.6.7-v0.1.6.7@docs-notebooks
  radical.entk         : 1.6.7
  radical.gtod         : 1.5.0
  radical.pilot        : 1.6.6-v1.6.6-16-g3a1f6dd97@feature-resource_separation
  radical.saga         : 1.6.6
  radical.utils        : 1.6.7```
```

I was working offline and this happened when I tried to import EnTK. I do not think we should require to have a connection to use EnTK.
```
ConnectionError                           Traceback (most recent call last)
/var/folders/7_/bp8lz2cs5ys7l6bssn4b8_mc0000gn/T/ipykernel_47934/4202695549.py in <module>
      8 import radical.utils as ru
      9 import radical.pilot as rp
---> 10 import radical.entk as re
     11 import radical.analytics as ra
     12 

~/.virtualenvs/analytics/lib/python3.8/site-packages/radical/entk/__init__.py in <module>
     31 version = version_short
     32 
---> 33 r = req.get("https://pypi.org/pypi/radical.entk/json")
     34 versions = r.json()["releases"].keys()
     35 last_version = list(versions)[-1]

~/.virtualenvs/analytics/lib/python3.8/site-packages/requests/api.py in get(url, params, **kwargs)
     74 
     75     kwargs.setdefault('allow_redirects', True)
---> 76     return request('get', url, params=params, **kwargs)
     77 
     78 

~/.virtualenvs/analytics/lib/python3.8/site-packages/requests/api.py in request(method, url, **kwargs)
     59     # cases, and look like a memory leak in others.
     60     with sessions.Session() as session:
---> 61         return session.request(method=method, url=url, **kwargs)
     62 
     63 

~/.virtualenvs/analytics/lib/python3.8/site-packages/requests/sessions.py in request(self, method, url, params, data, headers, cookies, files, auth, timeout, allow_redirects, proxies, hooks, stream, verify, cert, json)
    540         }
    541         send_kwargs.update(settings)
--> 542         resp = self.send(prep, **send_kwargs)
    543 
    544         return resp

~/.virtualenvs/analytics/lib/python3.8/site-packages/requests/sessions.py in send(self, request, **kwargs)
    653 
    654         # Send the request
--> 655         r = adapter.send(request, **kwargs)
    656 
    657         # Total elapsed time of the request (approximately)

[...]

PR #592, opened 2021-07-20
ignore pip version check when offline 

lee212:
#591 pip version check is disabled when ConnectionError is raised.

### DIFF
--- a/src/radical/entk/__init__.py
+++ b/src/radical/entk/__init__.py
@@ -35,7 +35,10 @@ try:
     versions = r.json()["releases"].keys()
     last_version = list(versions)[-1]
     if parse_version(version) < parse_version(last_version):
-        warnings.warn("WARNING: You are using radical.entk version %s, however version %s is available." % (version, last_version), UserWarning)
+        warnings.warn("WARNING: You are using radical.entk version %s, however"
+                " version %s is available." % (version, last_version),
+                UserWarning)
+
 except ConnectionError:
     pass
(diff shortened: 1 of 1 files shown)


andre-merzky (approved):
I agree this solves the problem.  However, maybe it would be better to silently ignore the condition?  It is not really an error to use RE offline, and I don't think it warrants a warning.  Either way, I leave this to your judgement.

andre-merzky:
@lee212 : please note that the linters fail - please fix before merging.  Also, several lines are too long (we have an 80 char limit).

lee212:
> I agree this solves the problem. However, maybe it would be better to silently ignore the condition? It is not really an error to use RE offline, and I don't think it warrants a warning. Either way, I leave this to your judgement.

I eliminated the warning at  236769ee23eb11dd867ef47b0c9ec07af16dc06c

lee212:
> @lee212 : please note that the linters fail - please fix before merging. Also, several lines are too long (we have an 80 char limit).

Running linter on PR diff is failing because of `E0401: Unable to import`, for example,` 'hypothesis' (import-error)`
The current dependency is here: https://github.com/radical-cybertools/radical.entk/blob/devel/.github/workflows/python-app.yml#L98-L104, do we want to add all packages under pylint actions? Otherwise, I merge this PR for now and close.

lee212:
I will merge this for now. I will fix pylint test if it complains again.

PR #597, opened 2021-11-19
ensure dir staging

andre-merzky:
this is needed for repex, but fixes the recursive staging for RE in general.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -16,7 +16,8 @@ import shutil
 
 import subprocess as sp
 
-from setuptools import setup, Command, find_packages
+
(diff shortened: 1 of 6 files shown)


mtitov:
This should be merged after a corresponding PR in RP, since currently RP doesn't have `rp.RECURSIVE`

andre-merzky:
This PR now also contains changes for clean termination on `post_exec` failures.  W/o those changes, RE will hang forever on such failures until eventually the sync thread will miss heartbeats and then terminate with an totally unrelated error message.

mtitov (approved):
LGTM

PR #600, opened 2022-02-05
Entk resource descr. memory

AymenFJA:
This PR is to allow `EnTK` to pass the memory option on the pilot level to `RP`.


mtitov:
@AymenFJA Aymen, also add check into `_validate_resource_desc` method and need to fix + extend tests (specifically to check the correctness of setting this attribute and its validation)

mtitov (approved):
LGTM

PR #601, opened 2022-02-17
Fix codecov

mtitov:
Fix tests coverage
Currently codecov checks coverage over `tests/test_component`, besides `src/radical/entk`


mtitov:
closed in favor #602

PR #605, opened 2022-02-27
Hotfix/no res check

andre-merzky:
Remove check which is now more efficiently implemented in RP


mtitov (approved):
LGTM!
just linting needs a fix

PR #609, opened 2022-04-03
Update setup requirements

mtitov:
Package `sphinx` is for documentation, but its dependencies as `packaging` and `requests` are used by EnTK, thus set these EnTK dependencies explicitly


andre-merzky (approved):
LGTM

PR #611, opened 2022-04-05
Fix/no uid on none

andre-merzky:
Attempt to fix that old, misguided `AttributeError: 'NoneType' object has no attribute '_uid'` error on abnormal RE termination.

### DIFF
--- a/src/radical/entk/utils/prof_utils.py
+++ b/src/radical/entk/utils/prof_utils.py
@@ -131,47 +131,52 @@ def write_session_description(amgr):
         'event_model': dict(),
     }
 
-    # Adding amgr to the tree
     tree = dict()
-    tree[amgr._uid] = {'uid': amgr._uid,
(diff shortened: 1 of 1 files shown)


andre-merzky:
> p.s. I've moved `amgr` check at the beginning of this func, since if it is not provided, everything else will be failed

I am afraid that changes semantics - previously the method still wrote the json file, only that the respective sessions were missing.  I would like to keep that behavior to avoid follow-up errors for the missing file.  Unless I am missing something, would you mind to revert that again?

mtitov:
as I understood, if `amgr` is not set, then it is not possible to get any associated attributes, thus json-file also will not be created (because it requires `amgr.sid, amgr.sid`) - and it helps to avoid raising `AttibuteError`

andre-merzky:
> since we do check here, but latter still access amgr without checking either it is set

Thanks, got it.

PR #615, opened 2022-07-27
Removed `tests/test_issues` (duplication)

mtitov:


.gitignore
```diff
@@ -68,20 +68,11 @@ docs/_build/
 
 # pytest files
 src/radical/entk/tests/__pycache__
-tests/__pycache__
-tests/test_api/__pycache__
-tests/test_functionality/__pycache__
-tests/test_integration/__pycache__
-tests/test_issues/__pycache__
-
+tests/*/__pycache__
```
andre-merzky, line 71: Should we use `**/__pycache__` - we also don't want that in other places...?

.gitignore
```diff
@@ -68,20 +68,11 @@ docs/_build/
...
 
 # hypothesis files
 src/radical/entk/tests/.hypothesis
-tests/.hypothesis
-tests/test_api/.hypothesis
-tests/test_functionality/.hypothesis
-tests/test_integration/.hypothesis
-tests/test_issues/.hypothesis
+tests/*/.hypothesis
```
andre-merzky, line 75: Similar, maybe `tests/**/.hypothesis`?

andre-merzky (approved):
LGTM, thanks!

.gitignore
mtitov: This was a duplication, but one that is there has a larger scope

.gitignore
mtitov: expanded (and there were duplications)

PR #617, opened 2022-09-01
Added timeout for `requests.get`

mtitov:


andre-merzky (approved):
I don't think this is a useful warning anyway...  Having said that, the timeout makes sense of course - thanks!

PR #618, opened 2022-09-19
Feature/task env

andre-merzky:
Is there a reason *not* to merge this?


mtitov:
@andre-merzky it was used for some use case, but wasn't sure we want to push further, but I'm ok to have it! p.s. just updated failed tests

mtitov (approved):

andre-merzky:
Great, thanks!

PR #620, opened 2022-11-05
inform RE about RP level task errors

andre-merzky:
This PR forwards exception information from RP to RE and let's RP report them.  This works sometimes.  ARGH!  I don't see how on some runs all tasks get the exception information, and on other identical runs they don't.  The RP layer reporting is the same every time - but somehow the RMQ state update messages sometimes are out of order, and old updates overwrite the setting of new ones.  What a mess...


mturilli:
This needs to be investigated and understood. We need to find a solution that, at the moment, we don't yet know. Sprint 20221223 will be used to understand the solution. If viable, to implement it too.

andre-merzky:
This is now ready for review.  The PR is a bit larger than the semantic changes warrant, but the added (and commented for now) log messages and the light refactoring make it easier to trace the task updates throughout the code.

mturilli:
Review will be long as it is a major PR.

mtitov (approved):
LGTM! 
and would suggest to remove commented lines used for debug (not used logging calls)

andre-merzky:
> would suggest to remove commented lines used for debug (not used logging calls)

I removed them now.  It took me a long while to understand the path of the task update messages, and those logs helped but were too verbose to leave active.  But we have them in git for the next time :-D

PR #626, opened 2022-12-13
remove RabbitMQ dependency from EnTK and replace with ZMQ communication.

andre-merzky:


src/radical/entk/appman/appmanager.py
```diff
@@ -111,7 +97,6 @@ def __init__(self,
         self._prof.prof('amgr_creat', uid=self._uid)
 
         self._rmgr            = None
-        self._pending_queue   = list()
         self._completed_queue = list()
```
mtitov, line 100: leftover, safe to remove

src/radical/entk/appman/wfprocessor.py
```diff
@@ -216,28 +224,12 @@ def reset_workflow(self):
...
-        wl_json = {'type': 'workload',
-                   'body': [task.as_dict() for task in workload]}
-
-        wl_json = json.dumps(wl_json)
-
-        # Acquire a connection+channel to the rmq server
-        mq_connection = pika.BlockingConnection(self._rmq_conn_params)
-        mq_channel = mq_connection.channel()
+        msg = {'type': 'workload',
+               'body': [task.as_dict() for task in workload]}
```
mtitov, line 228: general point about structure of zmq messages - some messages are of format `{'type':.., 'body':..}`, some are `{'type':.., 'object':..}`, and some are just of Task in a dictionary format - would propose to have one structure

mtitov (commented):
first set of comments

.github/workflows/python-app.yml
```diff
@@ -77,8 +73,7 @@ jobs:
         python -m pip install git+https://github.com/radical-cybertools/radical.saga.git@devel
         python -m pip install git+https://github.com/radical-cybertools/radical.pilot.git@devel
         python -m pip install --upgrade .
-        python -m pip install flake8 flake8-per-file-ignores
-        python -m pip install pytest pylint
+        python -m pip install -r requirements-tests.txt
```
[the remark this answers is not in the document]
andre-merzky: I think it's fine to keep test and linting requirements in one file - would be overkill to split IMHO

src/radical/entk/appman/appmanager.py
andre-merzky: Thanks, done.

src/radical/entk/appman/appmanager.py
```diff
@@ -535,102 +500,58 @@ def resource_terminate(self):
...
-        run, we first delete the queue and create a new one.
         '''
 
         try:
-            self._report.info('Setting up RabbitMQ system')
-            if self._mqs_setup:
+            sid = self._sid
+            self._report.info('Setting up ZMQ queues')
+
+            if self._zmq_setup:
```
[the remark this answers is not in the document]
andre-merzky: Right, I meant to test `self._zmq_info`, fixed now.
Fixed the leftover, thanks.

src/radical/entk/appman/appmanager.py
```diff
@@ -837,52 +713,51 @@ def _get_message_to_sync(self, mq_channel, qname):
...
-                self._update_task(msg, return_queue_name, props.correlation_id,
-                        mq_channel, method_frame)
+                self._update_task(obj)
 
 
     # --------------------------------------------------------------------------
     #
-    def _update_task(self, msg, reply_to, corr_id, mq_channel, method_frame):
+    def _update_task(self, tdict, reply_to=None, corr_id=None, mq_channel=None,
+                           method_frame=None):
```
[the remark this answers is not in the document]
andre-merzky: thanks, removed.

src/radical/entk/appman/appmanager.py
```diff
@@ -892,90 +767,52 @@ def _update_task(self, msg, reply_to, corr_id, mq_channel, method_frame):
...
+                        state = tdict['state']
                         self._prof.prof('pub_ack_state_%s' % state,
-                                        uid=msg['object']['uid'])
+                                        uid=tdict['uid'])
 
-                        mq_channel.basic_ack(
-                                delivery_tag=method_frame.delivery_tag)
+                        if mq_channel:
+                            mq_channel.basic_ack(
+                                    delivery_tag=method_frame.delivery_tag)
```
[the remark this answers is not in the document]
andre-merzky: Removed

src/radical/entk/appman/wfprocessor.py
andre-merzky: @mtitov: I started to do that, but realized that it does not make as much sense as I thought it would.  In RP, we have uniform message structured on the pubsub channels: any component can listen on those channels and then needs to figure out what message is 'interesting' and what message can be ignored.  We use the `cmd` field for that which every pubsub message has.  On the queues, we only expect one specific component to listen and that component expects a specific message type on that queue, so we don't have the same structure there.
Same here in EnTK: we use queues, and while the message types differ (task dicts vs. generic objects vs. resource requests), the recipient already knows what message type(s) to expect, and adding more structure adds just noise and code really.
In an abstract way I still agree somewhat: having better defined message structures would make the code easier to inspect for new developers for example, or would make logging on the ZMQ queues simpler as they could better inspect messages, but if we want to do that, then better in sync with RP.
Summary: out of scope for this PR :-)

.github/workflows/python-app.yml
```diff
@@ -36,11 +36,7 @@ jobs:
...
         python -m pip install --upgrade .
-        python -m pip install psutil
-        python -m pip install hypothesis
-        python -m pip install coverage codecov
-        python -m pip install pytest pytest-timeout pytest-xdist
-        python -m pip install timeout-decorator
+        python -m pip install -r requirements-tests.txt
     - name: Test with pytest
       env:
         RMQ_HOSTNAME: localhost
```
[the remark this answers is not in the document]
andre-merzky: Done.

src/radical/entk/appman/appmanager.py
```diff
@@ -111,11 +97,8 @@ def __init__(self,
         self._prof.prof('amgr_creat', uid=self._uid)
```
mtitov, line 97: a little offtop, but just to make reporting nicer
L96 -> `self._report.info('Creating AppManager\n')`
L123 -> 
```
        self._report.info('AppManager initialized')
        self._report.ok('>>ok\n')
```

src/radical/entk/appman/appmanager.py
```diff
@@ -67,16 +59,11 @@ class AppManager(object):
...
-                 password=None,
                  reattempts=None,
                  resubmit_failed=None,
                  autoterminate=None,
                  write_workflow=None,
                  rts=None,
-                 rmq_cleanup=None,
                  rts_config=None,
                  name=None,
                  base_path=None):
```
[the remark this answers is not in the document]
andre-merzky: Since we decided to opt for a major version number bump, I would rather opt not to do that - it would only mask incorrect use really...

src/radical/entk/appman/appmanager.py
andre-merzky: I am not exactly sure what you mean - but I am all for improving the reporter output.  Would you mind pushing a patch?

src/radical/entk/appman/appmanager.py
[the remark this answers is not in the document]
mtitov: we can add a warning about deprecation if any of the following keys are provided (`hostname`, `port`, `username`, `password`)

src/radical/entk/appman/appmanager.py
mtitov: that is a small update for not to have stick together two messages (`Creating AppManagerSetting up ZMQ queues`)
and to have something like this
```
Creating AppManager
Setting up ZMQ queues                                                         ok
AppManager initialized                                                        ok
Validating and assigning resource manager                                     ok
Setting up ZMQ queues                                                        n/a

```

andre-merzky: Ah, of course :-)

src/radical/entk/appman/appmanager.py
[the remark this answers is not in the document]
andre-merzky: This is implemented now - thanks.

src/radical/entk/appman/appmanager.py
andre-merzky: Done.

src/radical/entk/appman/appmanager.py
```diff
@@ -535,102 +505,58 @@ def resource_terminate(self):
...
+            self._logger.debug('Setting up zmq queues')
 
-            for q in qs:
-                # Durable Qs will not be lost if rabbitmq server crashes
-                mq_channel.queue_declare(queue=q)
+            cfg = ru.Config(cfg={'channel'   : sid,
+                                 'uid'       : sid,
+                                 'path'      : sid,
+                                 'type'      : 'queue',
+                                 'log_lvl'   : 'DEBUG',
```
mtitov, line 537: should we read log level from a corresponding env variable?

src/radical/entk/appman/appmanager.py
mtitov: tried it again,  and there is a line with lonely `ok`
```
EnTK session: re.session.b7da4236-9926-11ed-b1d0-0242c0a86005
Creating AppManager
Setting up ZMQ queues                                                         ok
                                                                              ok
```

let's add `self._report.info('AppManager initialized')` before [L130](https://github.com/radical-cybertools/radical.entk/blob/b877c876539b3dfa0fd4199cd95603114a20b013/src/radical/entk/appman/appmanager.py#L130)?

andre-merzky: Sure, done.

src/radical/entk/appman/appmanager.py
andre-merzky: You are right - I removed the fixed setting, the env will kick in from [here](https://github.com/radical-cybertools/radical.utils/blob/devel/src/radical/utils/zmq/bridge.py#L46) where the `cfg` setting will be `None` now.

PR #628, opened 2023-01-04
Sync with RP Task attributes `tags` - `exclusive`

mtitov:
This was required to distributed tasks (exclusively) among provided nodes for ExaAM workflows (ExaCA, OpenFOAM) on Crusher


src/radical/entk/execman/rp/task_processor.py
```diff
@@ -177,32 +177,19 @@ def resolve_arguments(args, placeholders, logger):
 #
 def resolve_tags(task, parent_pipeline_name, placeholders):
 
-    # entk only handles co_location tags.  If tags are given as strings, they
-    # get translated into `{'colocate': '<tag>'}`.  Tags passed as dictionaries
-    # are checked to conform with above form.
-    #
-    # In both cases, the tag string is expanded with the given placeholders.
+    tags = dict(task.tags or {'colocate': task.uid})
```
andre-merzky, line 180: Am I reading this right that the `colocate` tag is _*always*_ specified by default?  That's not part of the PR, but surprising nevertheless.  From the comment, I would have expected the code to look more like
```py
if isinstance(task.tags, str):
    tags = {'colocate': task.tags}
elif isinstance(task.tags, dict):
    tags = task.tags
else:
    tags = dict()
```
or something like that.  Since the `colocate` list's the task's _*own*_ uid, what is the expected scheduler semantics for that tag?

src/radical/entk/execman/rp/task_processor.py
```diff
@@ -177,32 +177,19 @@ def resolve_arguments(args, placeholders, logger):
...
-        if colo_tag in placeholders[parent_pipeline_name][sname]:
-            return {'colocate': placeholders[parent_pipeline_name][sname][colo_tag]['uid']}
-
-    for pname in placeholders:
-
-        # skip self pipeline this time
-        if pname == parent_pipeline_name:
-            continue
-
+    for pname in [parent_pipeline_name] + pnames:
```
andre-merzky, line 186: this ensures that the `parent_pipeline_name` is always included in the loop.  Are we sure that this is always in the `placeholders` dict?  Or should we add a check?  I tried to dig through the placeholder creation, but oh boy, that code is a challenge.  What's this `__by__name__ thingie?  Anyway, that's unrelated to this PR...

andre-merzky (approved):
This placeholder business is a bit of black magic - glad it works for the ExaAM use case!  While I left some comments, those are mostly caused by my ignorance, not by actual complains.

LGTM!

src/radical/entk/execman/rp/task_processor.py
mtitov: > I would have expected the code to look more like

yeah, that comments are confusing, that's why I removed them :) as of Task's description the type of `tags` is always `dict`, thus it is either provided as a dict or created here with task's `uid`

> what is the expected scheduler semantics for that tag?

since every task will have its own tag, then it is the same as no tasks have any tag. Honestly I didn't try to understand why tags were enforced..

src/radical/entk/task.py
```diff
@@ -527,7 +529,7 @@ def _post_verifier(self, k, v):
                         actual_value=_v)
 
         elif k == 'tags':
-            if list(v) != ['colocate']:
+            if any(tag_k not in VALID_TAG_KEYS for tag_k in v):
```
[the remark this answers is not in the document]
mtitov: I guess to make it clear we need to understand the meaning of a `tag`, I thought it is a value per a corresponding kv-pair, but probably we can refer to the whole kv-pair and its key in particularly. I'll update then

src/radical/entk/execman/rp/task_processor.py
mtitov: with this one I didn't change the logic, just tried to simplify it a little - to have one pass through all values and ensure that `parent_pipeline_name` comes the first. Same as above - also didn't try to dig further here

andre-merzky:
> also didn't try to dig further here
Yeah, that's fair, makes sense to keep that PR scoped as it is.  Thanks.

PR #630, opened 2023-01-13
Fix/mock

andre-merzky:
The Mock backend broke at some point.  This PR brings it back in sync with the RP backend.  The PR also contains cosmetic changes for the latter to minimize the diff (makes it easier to maintain both) and to get closer to our coding guidelines.


src/radical/entk/execman/rp/task_manager.py
```diff
@@ -279,6 +280,9 @@ def task_state_cb(rp_task, state, cb_data):
 
                     task = create_task_from_rp(rp_task, self._log, self._prof)
 
+                    import pprint
+                    pprint.pprint(task.as_dict())
```
mtitov, line 284: debug leftover, should be removed

mtitov (approved):
LGTM!
just some debug operations should be cleaned out before merge

mtitov:
@andre-merzky this branch got merged with no-rmq changes, lets push it towards `feature/no_rmq_2` branch then (which will be pushed toward `devel`). Otherwise there could be some inconsistency..

requirements.txt
```diff
@@ -1,7 +1,3 @@
 
-radical.utils>=1.12
```
mtitov, line 2: this still would be needed (used in `radical.entk.__init__` module)
p.s. and agree to remove others

andre-merzky: Well, it is pulled by RP anyway, thus the removal.  But I don't mind keeping it as explicit dependency anyway since it is directly imported...  Will re-add.

PR #631, opened 2023-01-14
Fix/mock no rmq

andre-merzky:
Same as #630 but targeting the `feature/no_rmq_2` branch.


requirements.txt
```diff
@@ -1,6 +0,0 @@
-
-radical.utils>=1.12
-radical.pilot>=1.18
-packaging
-requests>=2.4.0
```
mtitov, line 5: requirements files got deleted - that was accidentally, right?

examples/misc/scale.py
```diff
@@ -0,0 +1,74 @@
+#!/usr/bin/env python
+
+import os
+import sys
+import time
+
+from radical.entk import Pipeline, Stage, Task, AppManager
+
+host = os.environ.get('RMQ_HOSTNAME', 'localhost')
+port = os.environ.get('RMQ_PORT',     5672)
```
mtitov, line 10: these lines could be removed

examples/misc/scale.py
```diff
@@ -0,0 +1,74 @@
...
+
+                s.add_tasks(t)
+
+            p.add_stages(s)
+
+        pipelines.add(p)
+
+
+    appman = AppManager(autoterminate=True, rts='mock')
+    appman.resource_desc = {'resource': 'local.localhost',
```
mtitov, line 46: should we use another resource - `local.localhost_test`? which will not create a new VE, and will reuse the current one

requirements-tests.txt
```diff
@@ -2,15 +2,9 @@
...
-pytest-xdist
 pylint
 flake8
-flake8-per-file-ignores
-psutil
-mock==2.0.0.
-hypothesis
 coverage
-codecov
-timeout-decorator
```
mtitov, line 15: were that removed packages excessive?

andre-merzky: Thanks for catching this!  I messed something up with the requirements.  Well, it's a good opportunity to test if they are needed actually :-)  I re-added those which are actually used.

examples/misc/scale.py
andre-merzky: Done, thanks - and some more too...

examples/misc/scale.py
andre-merzky: Right, done.

requirements.txt
andre-merzky: Thanks, resolved.

mtitov (approved):
LGTM!

PR #636, opened 2023-02-17
Enable GPU sharing (convert `gpu_reqs.gpu_processes` from type `int` into `float`)

mtitov:
In sync with RP PR [#2832](https://github.com/radical-cybertools/radical.pilot/pull/2832)


andre-merzky (approved):
Thanks, LGTM

PR #642, opened 2023-04-27
introduced service task

mtitov:


mtitov:
@andre-merzky there is one thing which I don't know how to handle (in case if we need to): while creating `re.Task` object, `uid` is assigned, but on the side of RPAgent, corresponding service task obtains a different `uid` (https://github.com/radical-cybertools/radical.pilot/blob/devel/src/radical/pilot/agent/agent_0.py#L372). For now it might be not critical, but would be good to review it at some point.

andre-merzky:
> @andre-merzky there is one thing which I don't know how to handle (in case if we need to): while creating `re.Task` object, `uid` is assigned, but on the side of RPAgent, corresponding service task obtains a different `uid` (https://github.com/radical-cybertools/radical.pilot/blob/devel/src/radical/pilot/agent/agent_0.py#L372). For now it might be not critical, but would be good to review it at some point.

Fair - we probably should shield the uid generation and check if the description has a UID already, just like we do for other tasks also.

src/radical/entk/appman/appmanager.py
```diff
@@ -298,13 +316,36 @@ def resource_desc(self, value):
...
+            tasks = [tasks]
+
+        for task in tasks:
+            if not isinstance(task, Task):
+                raise EnTKTypeError(expected_type=Task,
+                                    actual_type=type(task))
+
+        self._services = tasks
+        if self._rmgr:
+            self._rmgr.services = tasks
```
andre-merzky, line 346: It's probably outside of the scope of this PR, but I want to mention it.  That pattern to keep the data as member attribute or as rmgr attribute, depending on the call order, is really confusing.  I think data should be kept as member attributes in general, and the rmgr should only be created on `self.run()`, and only then get assigned the attributes all at once.  That would simplify several attribute setters and also make error modes more predictable.

But, as said, out of scope for here IMHO...

tests/test_component/test_amgr.py
```diff
@@ -72,6 +71,22 @@ def test_amgr_read_config(self, mocked_init, d):
...
+        services = [re.Task()]
+        amgr.services = services
+        # without initialized ResourceManager
+        self.assertEqual(amgr._services, services)
+
+        amgr._rmgr = mock.Mock()
+        amgr.services = services
+        # with initialized ResourceManager
+        self.assertEqual(amgr._rmgr.services, services)
+
```
andre-merzky, line 89: The third option would be to attach services to the amgr, then attach resources to the amgr - then the rmgr should know the services also.

andre-merzky (approved):
Left a comment on test coverage, but otherwise looks complete and correct to me.  Thanks!

tests/test_component/test_amgr.py
mtitov: done

src/radical/entk/appman/appmanager.py
mtitov: agree, and I didn't want to touch that part at this PR

PR #643, opened 2023-05-03
added logging for task queue

mtitov:


mtitov:
@andre-merzky I suspect, that `queue.task_done()` wasn't always applied and that could trigger EnTK to freeze (not to receive tasks). I couldn't reproduce such issue, but this small refactoring for queue tasks handling and extra logging would help to track it (latest runs on Frontier with this branch went successfully)

src/radical/entk/execman/rp/task_manager.py
```diff
@@ -267,19 +269,18 @@ def task_state_cb(rp_task, state):
...
-                body = None
-
                 try:
                     body = task_queue.get(block=True, timeout=10)
-
                 except queue.Empty:
                     # Ignore, we don't always have new tasks to run
-                    pass
-
-                if not body:
```
andre-merzky, line 279: IIRC, we met sometime conditions where the above `get` did not raise but the `body` was still `None`.  Thta was with RMQ though.  But I don't know if that was an RMQ artifact or if, in some cases, a `None` was pushed into the queue.  So it might be safer to keep that line, maybe.  Or maybe not, and we track this down when it happens again :-)

andre-merzky (approved):
LGTM

src/radical/entk/execman/rp/task_manager.py
mtitov: the line `if not body:` is kept, here is just small rearrangement to ensure that `task_done` is called whenever anything comes in. And yeah, keeping an eye on this!)

PR #644, opened 2023-05-10
fixed reattempts and sizing for `task_queue`

mtitov:


andre-merzky (approved):
Thanks @mtitov - this looks very reasonable to me.  It opens a small race condition in case some thread operates on the `session._dbs` after the `mongodb_close` or before the `mongodb_start` - but I don't see an immediate candidate in RE or RP to hit that race.  If it becomes a problem at some point, you may want to check `radical.utils.atfork` for a thread safe approach - but that in turn requires global variables and is such syntactically not a nice as your current code.

PR #650, opened 2023-07-17
Create workflow graph

mtitov:
Enables workflow graph creation (later will be expanded towards provenance graph)


src/radical/entk/utils/provenance.py
```diff
@@ -0,0 +1,82 @@
...
+        for stage in pipelines.stages:
+            graph[pipeline.uid][stage.uid] = []
+
+            for task in stage.tasks:
+                graph[pipeline.uid][stage.uid].append(task.uid)
+
+            graph[pipeline.uid][stage.uid].sort()
+
+    if output_file:
+        if '/' not in output_file:
```
andre-merzky, line 36: Why do you want to enforce a `/` in the file name?

src/radical/entk/utils/provenance.py
```diff
@@ -0,0 +1,82 @@
...
+            setdefault(pipeline_uid, {}).\
+            setdefault(stage_uid, []).\
+            append(task_uid)
+
+    for pipeline_uid in graph:
+        for stage_uid in graph[pipeline_uid]:
+            graph[pipeline_uid][stage_uid].sort()
+
+    if output_file:
+        if '/' not in output_file:
```
andre-merzky, line 72: same

andre-merzky (approved):
LGTM

src/radical/entk/utils/provenance.py
mtitov: thought to be explicit where file is saved. Should I remove this check? (and only to keep `.json` check)

andre-merzky: Yeah, I would suggest to remove.  Thanks!

PR #651, opened 2023-07-26
added possibility to annotate dataflow

mtitov:
based on previous PR https://github.com/radical-cybertools/radical.entk/pull/650


mtitov:
@mturilli @andre-merzky I've responded to Polina's comments and PR is ready to be reviewed/merged

src/radical/entk/pipeline.py
```diff
@@ -406,9 +406,29 @@ def _validate(self):
...
+                if not task.annotations:
+                    continue
+
+                annotated_outputs[task.uid] = task.annotations.outputs
+                for t_input in task.annotations.inputs:
+                    if ':' not in t_input:
+                        continue
+                    # ensure that task inputs are annotated as outputs
+                    # for corresponding tasks it depends on
+                    t_dep_uid, t_dep_output = t_input.split(':')
```
mturilli, line 425: Catch only the first part

mtitov: @mturilli I've extended this coverage, ready for the next round review/merge

src/radical/entk/pipeline.py
```diff
@@ -406,9 +406,33 @@ def _validate(self):
...
+                if not task.annotations:
+                    continue
+
+                annotated_outputs[task.uid] = task.annotations.outputs
+                for t_input in task.annotations.inputs:
+                    delimiter_count = t_input.count(':')
+                    if not delimiter_count:
+                        continue
+                    elif delimiter_count == 1 and '://' in t_input:
+                        # extend file representation format (future feature)
```
mturilli, line 425: Should we mark the comment with `TODO`, i.e., `# TODO: extend file representation format`? In that way, we would have a 'standardized' way to annotate future TODOs, and many IDEs can pick that up automatically.

src/radical/entk/task.py
```diff
@@ -586,6 +610,50 @@ def luid(self):
...
+    def annotate(self, inputs: Union[Dict, List, str, None] = None,
+                       outputs: Union[List, str, None] = None) -> None:
+        """
+        Adds dataflow annotations with provided input and output files,
+        and defines dependencies between tasks.
+
+        Attributes:
+            inputs (list, optional): List of input files. If file is produced
+                by previously executed task, then the corresponding input
+                element is provided as a dictionary with task instance as a key.
```
mturilli, line 624: "If file is produced by previously executed task, then the corresponding input element is provided as a dictionary with task instance as a key" -> If a file is produced by the previously executed task, then the corresponding input element is provided as a dictionary with the task instance as a key?

tests/test_component/test_task.py
```diff
@@ -422,6 +424,38 @@ def test_eq_hash(self):
...
+
+        t1 = Task()
+        t2 = Task()
+
+        t1.annotate()
+        self.assertIsInstance(t1.annotations.inputs,     list)
+        self.assertIsInstance(t1.annotations.outputs,    list)
+        self.assertIsInstance(t1.annotations.depends_on, list)
+
+        t1.annotate(inputs=['file_1.txt', 'file_2.txt'],
```
mturilli, line 439: Based on the code, having file names with [._0-9a-z] characters should not be an issue but, more in general, maybe worth investigating if we have a module we can use to test file names with a variety of non alphanumericals? Bugs related to characters in names bitten us in the past.

mturilli (commented):
Left some very minor comments. Once you consider them, IMHO it is ready to go.

src/radical/entk/task.py
mtitov: done

src/radical/entk/pipeline.py
mtitov: done

tests/test_component/test_task.py
mtitov: yeah, makes sense to look for corresponding libraries, but as part of other PR (I would think, we can focus on it after having similar functionality with RP)

mturilli (approved):
LGTM

PR #652, opened 2023-09-25
Devel nodb 2

mtitov:


mtitov:
FIXME: after session is finished, the result file `tmgr.0000.json` doesn't contain info about tasks

mturilli (approved):
As discussed, I agree it is not pretty but as it works, I think for now we can merge it.

mturilli:
Looking at the tests, should we rapidly address the obsolete keys? Also, following what discussed for RP, should we enable profiles by default?

mtitov:
> Looking at the tests, should we rapidly address the obsolete keys?

@mturilli this tests fail due to having RU,RP-devel branches without NoDB, that will be resolved after merging RU and RP into devel (will have some delay in merging EnTK, thus to confirm that tests pass)

> Also, following what discussed for RP, should we enable profiles by default?

it is enabled by default now

PR #654, opened 2023-09-29
fixed task manager handling

mtitov:
With this fix we don't share TaskManager among different processes, it will be created and handled within child process only (and information about tasks is in one place - corresponding output JSON file contains tasks info, which was missing earlier)


andre-merzky (approved):
Thanks for spotting this!

PR #661, opened 2024-02-27
Fix/darshan

mtitov:
Updated Darshan integration based on the latest experience of using it

### DIFF
--- a/src/radical/entk/tools/darshan.py
+++ b/src/radical/entk/tools/darshan.py
@@ -50,10 +50,10 @@ def cache_darshan_env(darshan_runtime_root: Optional[str] = None,
 
 # ------------------------------------------------------------------------------
 # decorator to enable darshan for function that generates Pipeline, Stage, Task
-def darshan(func,
-            darshan_runtime_root: Optional[str] = None,
-            modules: Optional[List[str]] = None,
-            env: Optional[Dict[str, str]] = None):
+def with_darshan(func,
+                 darshan_runtime_root: Optional[str] = None,
+                 modules: Optional[List[str]] = None,
+                 env: Optional[Dict[str, str]] = None):
     def wrapper(*args, **kwargs):
         return enable_darshan(func(*args, **kwargs),
                               darshan_runtime_root=darshan_runtime_root,
--- a/src/radical/entk/tools/darshan.py
+++ b/src/radical/entk/tools/darshan.py
@@ -64,14 +64,18 @@ def darshan(func,
 
 # ------------------------------------------------------------------------------
 #
-def enable_darshan(pst_obj: Union[Pipeline, Stage, Task],
+def enable_darshan(pst: Union[Pipeline, Stage, Task, List[Pipeline]],
                    darshan_runtime_root: Optional[str] = None,
                    modules: Optional[List[str]] = None,
                    env: Optional[Dict[str, str]] = None
                    ) -> Union[Pipeline, Stage, Task]:
 
-    if not isinstance(pst_obj, (Pipeline, Stage, Task)):
-        raise TypeError('Provide PST object to enable Darshan')
+    if not pst:
+        raise ValueError('PST object is not provided')
+    elif isinstance(pst, list) and not isinstance(pst[0], Pipeline):
+        raise TypeError('List of Pipelines is not provided')
+    elif not isinstance(pst, (Pipeline, Stage, Task)):
+        raise TypeError('Non PST object provided')
 
     cache_darshan_env(darshan_runtime_root, modules, env)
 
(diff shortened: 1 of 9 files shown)


mtitov:
@andre-merzky ready for review

src/radical/entk/tools/darshan.py
```diff
@@ -0,0 +1,173 @@
...
+# ------------------------------------------------------------------------------
+#
+def enable_darshan(pst: Union[Pipeline, Stage, Task, List[Pipeline]],
+                   darshan_runtime_root: Optional[str] = None,
+                   modules: Optional[List[str]] = None,
+                   env: Optional[Dict[str, str]] = None
+                   ) -> Union[Pipeline, Stage, Task]:
+
+    if not pst:
+        raise ValueError('PST object is not provided')
```
andre-merzky, line 74: `pst` is not an optional parameter, so this error should never happen?

src/radical/entk/tools/darshan.py
```diff
@@ -0,0 +1,173 @@
...
+            return
+
+        darshan_log_dir = DARSHAN_LOG_DIR % {'sandbox': '${RP_TASK_SANDBOX}'}
+        darshan_enable  = (f'LD_PRELOAD="{_darshan_runtime_root}'
+                           '/lib/libdarshan.so" ')
+
+        if src_task.cpu_reqs.cpu_processes == 1:
+            darshan_enable += 'DARSHAN_ENABLE_NONMPI=1 '
+
+        src_task.executable  = darshan_enable + src_task.executable
```
andre-merzky, line 98: The `executable` should not be a shell command line IMHO.  I would suggest to morph
```py
{  
    'executable': 'foo', 
    'arguments': ['-b', 'bar']
}
```
into
```py
{  
    'executable': '/bin/sh', 
    'arguments': ['-c', 'LD_PRELOAD=/tmp/lib/libdarshan.so foo -b bar']
}
```

src/radical/entk/tools/darshan.py
```diff
@@ -0,0 +1,173 @@
...
+        src_task.pre_launch += [f'mkdir -p {darshan_log_dir}']
+        src_task.pre_exec.extend(
+            _darshan_activation_cmds +
+            [f'export DARSHAN_LOG_DIR_PATH={darshan_log_dir}'])
```
andre-merzky, line 102: This could (should?) got to the task environment?

src/radical/entk/tools/darshan.py
```diff
@@ -0,0 +1,173 @@
...
+    return pst
+
+
+# ------------------------------------------------------------------------------
+#
+def get_parsed_data(log: str, target_counters: Union[str, List[str]]) -> set:
+
+    data = set()
+
+    grep_patterns = '-e ' + ' -e '.join(ru.as_list(target_counters))
```
andre-merzky, line 131: Are we sure that `target_counters` is never an empty list?

src/radical/entk/tools/darshan.py
```diff
@@ -0,0 +1,173 @@
...
+    parser_cmd    = (f'darshan-parser {log} | grep {grep_patterns} | '
+                     "awk '{print $5\":\"$6}'")
+    out, err, ret = ru.sh_callout(parser_cmd, env=_darshan_env, shell=True)
+    if ret:
+        print(f'[ERROR] Darshan not able to parse "{log}": {err}')
+    else:
+        for o in out.split('\n'):
```
andre-merzky, line 138: Nitpick, but I would suggest `for line in ...`

src/radical/entk/tools/darshan.py
```diff
@@ -0,0 +1,173 @@
...
+            if not o:
+                continue
+            value, file = o.split(':')
```
andre-merzky, line 141: `o.split(':', 1)`

andre-merzky (approved):
Thanks @mtitov, the code looks reasonable to me.

What I am missing is some explanation what Darshan actually is and what it is used for.  For example: some methods are public methods and will show up in RTD, but don't have a docstring, so the user is left guessing what this is about.

src/radical/entk/tools/darshan.py
mtitov: this is just to catch empty (e.g., `[]`, `{}`, `''`) or `None` values

src/radical/entk/tools/darshan.py
mtitov: yes, thus each task would have its own path to logs

src/radical/entk/tools/darshan.py
mtitov: updated accordingly

src/radical/entk/tools/darshan.py
mtitov: updated

src/radical/entk/tools/darshan.py
mtitov: updated

src/radical/entk/tools/darshan.py
mtitov: would it be ok to have it within other PR? (I'll convert this comment into task and then will response with a corresponding PR)

p.s. I need to work on the next approach using Darshan config file, and some of the settings might go into it, and I'll keep that in sync with this

andre-merzky: Sure, that works!

PR #663, opened 2024-04-10
use darshan config file

mtitov:


andre-merzky (approved):
This looks all reasonable to me.

PR #670, opened 2024-07-24
ensure that failed tasks fail the workload - if so configured

### LINKED ISSUE #669: task failures are not popping up
A task which ultimately (after retries etc) fails should cause the stage to fail, which should cause the pipeline to fail, which should cause the app manager to abort waiting for completion.  None of this is happening.  To keep the API backward compatible, tasks should get a `error_is_fatal` flag which defaults to `False`, reproducing the current behavior.  We need to add code paths to handle that flag when set to `True`.

andre-merzky:
This fixes #669

### DIFF
--- a/examples/simple/bag_of_pipelines.py
+++ b/examples/simple/bag_of_pipelines.py
@@ -1,0 +1,102 @@
+#!/usr/bin/env python3
+
+import sys
+import time
+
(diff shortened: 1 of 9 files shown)


mtitov (approved):
LGTM!

